PR #6, opened 2016-02-16
scheme to be derived

davidbirdsong:
- allows a TLS offload agent to forward the scheme


davidbirdsong:
`opts.redirect_uri_scheme` takes precendence

PR #9, opened 2016-04-24
Added option client_authentication_mode.

therealcmj:
The current code uses client_secret_post - sending the client ID and secret in the POST body when communicating with the OP. I needed to test with an OP that only supports client_secret_basic so I added a new config option to allow for either mechanism to be used.

To do that I needed to pass the config changes into openidc_call_token_endpoint - I wound up passing "opts" in as a new first argument. That change allowed me to remove the 3rd argument (ssl_verify) argument since that value is already in opts.

If client_authentication_mode option is omitted in nginx.conf then my code will use the first value in the metadata.

Please let me know if you want me to do this in a different way.

### DIFF
--- a/lib/resty/openidc.lua
+++ b/lib/resty/openidc.lua
@@ -201,14 +201,14 @@ local function openidc_parse_json_response(response)
 end
 
 -- make a call to the token endpoint
-local function openidc_call_token_endpoint(opts, endpoint, body, ssl_verify)
+local function openidc_call_token_endpoint(opts, endpoint, body)
 
   ngx.log(ngx.DEBUG, "request body for token endpoint call: ", ngx.encode_args(body))
 
   local requestHeaders = {
       ["Content-Type"] = "application/x-www-form-urlencoded"
   }
-  if opts.client_authentication_mode == "client_secret_basic" then
+  if opts.token_endpoint_auth_method == "client_secret_basic" then
     requestHeaders.Authorization = "Basic "..ngx.encode_base64( opts.client_id..":"..opts.client_secret)
     ngx.log(ngx.DEBUG,"Authorization header '"..requestHeaders.Authorization.."'")
   end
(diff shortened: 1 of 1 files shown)


zandbelt:
thanks, I like it :-), a few remarks though:

1. the `ssl_verify` parameter in the `openidc_call_token_endpoint` function signature is now obsolete
2. I prefer the `opts` name `token_endpoint_auth_method` instead of `client_authentication_mode` to stay in line with the metadata element name
3. I don't think you handle the case where neither `client_authentication_mode` is set nor `token_endpoint_auth_methods_supported` is set in metadata (null pointer exception?)

if you could take care if this I'd be happy to pull

therealcmj:
All three suggestions included. I decided to default to client_secret_basic since the spec says that should be the default.

zandbelt:
thanks for this; I may refactor it in due time so that client authentication creds & method are passed in as parameters to the `openidc_call_token_endpoint` method explicitly and it is reusable across both token endpoint requests as well as token introspection

therealcmj:
Understood. If we're talking about futures then we'll probably also have to adjust the code to restrict values to either post or basic since that's all we've implemented.
Thanks for accepting the changes.

zandbelt:
I refactored the code in 43e8e18efff68e4a820cc3c21b797429dbc785d7; it would be nice if you could confirm that this works for you as well.

therealcmj:
I didn't do extensive testing but a quick smoke test shows that it works. Here's the relevant portion of the log:
`2016/05/10 07:41:36 [debug] 14890#0: *2 [lua] openidc.lua:417: openidc_get_token_auth_method(): no configuration setting for option so select the first method specified by the OP: client_secret_basic
2016/05/10 07:41:36 [debug] 14890#0: *2 [lua] openidc.lua:428: openidc_get_token_auth_method(): token_endpoint_auth_method result set to client_secret_basic`

zandbelt:
I'll take it for now ;-) thanks!

PR #11, opened 2016-06-08
Trying to fix session locking problems on some server side storages

bungle:
- Removed unneccessary initializers
- added some locals
- fixed some global variable leaks
- adjusted session to only start when it is modified (in hopes to fix locking problems)

### DIFF
--- a/lib/resty/openidc.lua
+++ b/lib/resty/openidc.lua
@@ -334,12 +334,11 @@ local function openidc_authorization_response(opts, session)
     return nil, err, session.data.original_url
   end
 
+  -- call the user info endpoint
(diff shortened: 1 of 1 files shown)


spauljon:
so I 
a) merged (to our fork that has minor customizations) up to commit 3f115c8 (which I think is enough), and
b) reverted the lua-resty-session change (to memcache.lua)

and tested it again (have a netscaler lb fronting two nginx servers supported by two memcache instances); at first glance, everything seems to work - no hanging

as an additional measure, I instrumented the lock and unlock code, and confirmed that the calls now are paired

so looks good ... thx for the timely response - we appreciate it

PR #13, opened 2016-07-08
Add Travis build

asbjornu:
This PR adds a [Travis build](https://travis-ci.org/asbjornu/lua-resty-openidc) to the project so the `.rockspec` is automatically built and uploaded to [LuaRocks](https://luarocks.org/modules/asbjornu/lua-resty-openidc) on every commit. Before merging this:

- @ping-identity's [Travis build for this repository](https://travis-ci.org/pingidentity/lua-resty-openidc) should be enabled. The one I link to above is my personal Travis project.
- Someone affiliated with @ping-identity should create a LuaRocks profile and get an API key from [the settings page](https://luarocks.org/settings/api-keys), so the rock is uploaded to a more "official" profile. The currently uploaded module belongs to my personal account, which isn't ideal.
- This API key must be configured in the [settings of the Travis project](https://travis-ci.org/pingidentity/lua-resty-openidc/settings) with the name `API_KEY` so it is made available to the `.travis.yml` file.

There's only one snag about this, which is that the `luarocks upload` command will fail if the package already exists for that version number. Ideally, every commit on the `master` branch should lead to a new version, and we could get this as a variable from `git tag` instead of hard coded in `.travis.yml` like now, but that requires some discipline on the maintainers of this project that I don't know if can be mandated.

Thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ install:
 
 script:
   - luarocks pack $ROCKSPEC
+  - luarocks upload $ROCKSPEC --api-key=$API_KEY # The API key comes from an environment variable configured on https://travis-ci.org/
 
 notifications:
   email:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,33 @@
+language: c
+
+sudo: false
+
+env:
+  global:
+    - NAME=lua-resty-openidc
+    - VERSION=1.1-0
+    - LUAROCKS=2.3.0
+  matrix:
+    - LUA=lua5.1
+
+before_install:
+  - source .travis/setenv_lua.sh
+  - luarocks install Lua-cURL --server=https://luarocks.org/dev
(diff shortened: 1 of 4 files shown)


zandbelt:
I'm not familiar with Travis but I'll look in to it. Wrt. to the master commits: isn't there an option to make Travis watch a different branch, says "releases"?

asbjornu:
@zandbelt A suggestion would be to use a more standard naming, like "develop" to have for unstable development and then merge that to master for every release and tag that merge with the version number. Alternatively, use "GitHubFlow" where every new feature has its own "feature branch" and merge that to master when it's done, then tag those merges with a version number. Common to most of these flows is that there's no regular commits directly on the master branch, only merge commits.

With a process like this, it's easy to tell Travis to behave differently based on the branch that is being built, with the added benefit of being a de facto flow people might be familiar with.

Issue #15, opened 2016-07-17
Upload Docker image to Docker Hub


asbjornu:
To be able to upload a Docker image of `lua-resty-openidc`, an organization and repository need to be set up on [Docker Hub](https://hub.docker.com/). When that's done, we should add a [`docker push`](https://docs.docker.com/v1.11/engine/userguide/containers/dockerrepos/#pushing-a-repository-to-docker-hub) step to the `after_deploy` section of the `.travis.yml` file, so a new Docker image is uploaded for every new tag that is built.

Issue #14, opened 2016-07-17
Add integration tests


asbjornu:
It would be nice if some sort of integration tests were easily available in this repository, either as something created for `lua-resty-openidc` itself, as a Git submodule (if tests exist elsewhere) or just as a reference in the form of a link in the README.

I've been searching for some sort of official test suite for OAuth and OpenID Connect, but the only thing I've found is [this](https://github.com/oauth-io/testing-oauth), which requires setting up all kinds of user-dependent stuff on [oauth.io](https://oauth.io/) and LinkedIn. It also didn't work after I went through the required steps, so I don't know if this is an achievable goal at all.

asbjornu:
@zandbelt This issue wasn't really related to #16 at all. We should have integration tests in this repository regardless of how we bootstrap them, through Docker or otherwise. If you have ideas on where to find a set of tests we can exercise against lua-resty-openidc, it would be great if you could share them.

PR #16, opened 2016-07-17
Add Dockerfile

asbjornu:
This PR does a few things, namely:

1. Adds a `Dockerfile` based on [`openresty/openresty:trusty`](https://hub.docker.com/r/openresty/openresty/), which makes it easier to get `lua-resty-openidc` up and running with all dependencies.
2. Removes the need to have a hard coded version number in the Lua `.rockspec` file; the version information is retrieved from `git describe` instead.
3. When a tag is being built, the packed `.rock` file will be uploaded to the repository's release for the given tag.
4. With the release tracking in place, uploading of the `.rock` to Lua Rocks has been moved to `after_deploy`, which means it won't be attempted for all builds; only for tagged ones.

A note to point 3 above: This requires a [personal access token](https://github.com/settings/tokens) to be created and added in [the Travis settings for the repository](https://travis-ci.org/pingidentity/lua-resty-openidc/settings) with the environment variable name `GITHUB_OAUTH_TOKEN`.

What's blatantly missing from this PR are tests to exercise the Docker container's NGINX server with OAuth and OpenID Connect test scenarios. I've added #14 to track this.

I'd also like to add uploading the Docker image to [Docker Hub](https://hub.docker.com/) with [`docker push`](https://docs.docker.com/v1.11/engine/userguide/containers/dockerrepos/#pushing-a-repository-to-docker-hub), but that requires quite a bit of setup on Docker Hub first. I've added #15 to track this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /.project
+*.rock
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,11 +40,6 @@ script:
   # TODO: Add some integration tests that excersice the OAuth and OpenID Connect
   # interface of lua-resty-openidc.
 
-after_success:
-  # The Lua Rocks API key comes from an environment variable configured on
-  # https://travis-ci.org/
-  - luarocks upload $ROCKSPEC --api-key=$API_KEY
-
 deploy:
   provider: releases
   # The GitHub OAuth Token comes from an environment variable configured on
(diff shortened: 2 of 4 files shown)


zandbelt:
I think this deserves to be in a separate repository on its own title, not even a fork of lua-resty-openidc. IMHO this repo should be core dev, any tooling around it could go elsewhere. That makes it easier to maintain as well since focus would be specific. Let me know what you think.

asbjornu:
@zandbelt That's a good idea, but it would make continuous integration more difficult, since we then need to (or at least should) trigger a build of the Docker image repository every time we change something in lua-resty-openidc, to ensure that the Docker image doesn't break unintentionally. I'm not sure how to do that in Travis, but can take a look if you still think that's the best way forward.

PR #24, opened 2016-10-18
Feature jwt discovery validation

hanikesn:
This patch adds easy verification of jwt tokens, via the JWKS from the discovery endpoint.

Please feel free to leave any suggestions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,66 @@ lAc5Csj0o5Q+oEhPUAVBIF07m4rd0OvAVPOCQ2NJhQSL1oWASbf+fg==
 }
 ```
 
+## Sample Configuration for OAuth 2.0 JWT Token Validation
+
+Sample `nginx.conf` configuration for verifying Bearer JWT Access Tokens against a OpenID Connect Discovery endpoint.
+Once successfully verified, the NGINX server may function as a reverse proxy to an internal origin server.
+
+```
+events {
+  worker_connections 128;
+}
+
+http {
+
+  lua_package_path '~/lua/?.lua;;';
+
+  resolver 8.8.8.8;
+  
+  # cache for JWT verification results
+  lua_shared_dict introspection 10m;
+  # cache for jwks metadata documents
+  lua_shared_dict discovery 1m;
+ 
+  server {
+    listen 8080;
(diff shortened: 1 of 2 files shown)


zandbelt:
this code is nice to have since it can be re-used for several scenario's, however:

1. technically Discovery is part of OpenID Connect and not (at least not yet) part of OAuth 2.0; since validation of access_tokens in the Resource Server role is pure OAuth and not OpenID Connect related, I would be interested to know which Authorization Servers already support Discovery + jwks_uri in this "pre-standard" way

2. I believe that most RSA JWKs (at least on OIDC jwks_uri's today) would be represented using the `n\e` compact syntax as in https://tools.ietf.org/html/rfc7517#appendix-A.1 instead of using `x5c`; any chance you'd be willing to contribute support for that to `lua-resty-jwt`...  :-0 ?

hanikesn:
> technically Discovery is part of OpenID Connect and not (at least not yet) part of OAuth 2.0; since validation of access_tokens in the Resource Server role is pure OAuth and not OpenID Connect related, I would be interested to know which Authorization Servers already support Discovery + jwks_uri in this "pre-standard" way

We're probably using it in a non-standard way to validate access_tokens which were created by the IDP in the first place.

> I believe that most RSA JWKs (at least on OIDC jwks_uri's today) would be represented using the n\e compact syntax as in https://tools.ietf.org/html/rfc7517#appendix-A.1 instead of using x5c; any chance you'd be willing to contribute support for that to lua-resty-jwt... :-0 ?

I guess we're really fortunate that [Azure Active Directory](https://login.microsoftonline.com/common/.well-known/openid-configuration) supports x5c. But I saw that e.g. Google does not. Maybe I can come up with a solution, it's just that OpenSSL isn't exactly easy to work with.

hanikesn:
I took a closer look at supporting modulus and exponent format directly and it would basically mean writing a small ASN.1 DER encoder for:

```
RSAPublicKey ::= SEQUENCE {
    modulus           INTEGER,  -- n
    publicExponent    INTEGER   -- e
}
```

As the Integers quite big some kind of big int support is probably also needed.

Issue #28, opened 2016-11-17
Access to un-parsed id_token JWT for Kubernetes OIDC integration


cllunsford:
I'm trying to use lua-resty-openidc as an OIDC RP in front of Kubernetes.  Based on the Kuberentes docs, it looks like the oidc integration requires the original jwt `id_token` instead of the `access_token`.  (https://github.com/kubernetes/kubernetes.github.io/pull/1209).  Would there be any issues with adding the unparsed `json.id_token` to the session data so I can set it in the Authorization Header when proxying to k8s?

Thanks!

zandbelt:
I don't see any security issues with that as the NGINX proxy and Kubernetes can be seen as part of the same system/client/domain. I'd be happy to merge a PR. Seems that it only requires:
https://github.com/gdestuynder/lua-resty-openidc/commit/88f5a0b55089cab92a9eb936ebb85085c48a0c83

cllunsford:
Great, thanks!  I've opened a PR with this change.

PR #29, opened 2016-11-18
Saving encrypted id_token to session data.

cllunsford:
This change allows the jwt id_token to be set on the proxied request.  Kubernetes OIDC integration requires the bearer token be set to the jwt id_token instead of the access_token.


cllunsford:
Closes #28

PR #47, opened 2017-04-21
Support for prompt parameter during authorization, returning session data, and logout url override

gdestuynder:
Hi,

these are 3 patches that we use at Mozilla (i.e. for Mozilla properties authenticating with OpenID Connect) with OpenID Connect. Let me know if you'd prefer 3 PRs, though some depend on each other.

The prompt parameter allows for silent re authentication in particular in this case (though this can be further extended later to support all options), as seen on http://testrp.security.allizom.org/. This allows to refresh user profile information without login the user out and back in. Spec at http://openid.net/specs/openid-connect-implicit-1_0.html#RequestParameters

The session data being returned allows for the calling code to get full session data (lua-resty-session does not always allow this otherwise)

Finally, the logout URL redirect allows to override the discovered redirect from the OP - or to set one if there was none discovered - particularly helpful when using through an OpenIDConnect access proxy.

Thanks for your consideration

### DIFF
--- a/lib/resty/openidc.lua
+++ b/lib/resty/openidc.lua
@@ -172,7 +172,8 @@ local function openidc_authorize(opts, session, target_url)
     scope=opts.scope and opts.scope or "openid email profile",
     redirect_uri=openidc_get_redirect_uri(opts),
     state=state,
-    nonce=nonce
+    nonce=nonce,
+    prompt=opts.prompt and opts.prompt or ""
   }
 
   -- merge any provided extra parameters
@@ -185,6 +186,7 @@ local function openidc_authorize(opts, session, target_url)
   session.data.original_url = target_url
   session.data.state = state
   session.data.nonce = nonce
+  session.data.last_authenticated = ngx.time()
   session:save()
 
   -- redirect to the /authorization endpoint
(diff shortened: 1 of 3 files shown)


lib/resty/openidc.lua
```diff
@@ -576,6 +578,13 @@ function openidc.authenticate(opts, target_url)
     return openidc_authorize(opts, session, target_url), session
   end
 
+  -- silently reauthenticate if necessary (mainly used for session refresh/getting updated id_token data)
+  if opts.refresh_session_internal and (session.data.last_authenticated == nil or (session.data.last_authenticated+reauth_delay) < ngx.time()) then
```
zandbelt: Seems that `opts.refresh_session_internal ` and `reauth_delay` are used without being defined? An alternative may be to define an optional custom callback function in `opts` that implements the requested behavior.

gdestuynder:
hi @zandbelt - my apologies, I pushed the wrong set of commits.. :(
To make this easier to overwrote my branch with what we *actually* use so that it does not open a new PR.

Note that this does not use a callback in opts though. One particular thing is that if you do not want to use `prompt=none` ("silent" authentication) the other options do not necessarily work (such as `prompt=login`) depending on the OIDC Provider (e.g. Auth0 OP supports `none` but not `login` or `consent` or `select_account`) - and the OIDC ref/spec is unclear if this is OK or not.

In doubt, I prefer to make it completely optional so that one can choose not to use `prompt` at all (default). In our use case, using `prompt=silent` for checking if the session is still valid is great though - and I believe most that can support that will want it. It allows you to keep users logged in as long as you want, since new user attributes/claims are passed over regularly, and this will also indicate when the user should be logged out)

For the callback - if you would like this implemented and have an example (I'm new to Lua, so that would help) I can do this of course

Thanks !

gdestuynder:
I added 2 commits which hopefully address your comment :)

Note: I think 15min is a useable value - this only occurs when a user is actively browsing the site. For example, if you disable the user (for any reason) from the OP it means that the access to the site/RP behind the access-proxy using lua-resty-openidc will stop serving that user within 15min. Regardless, made the whole thing optional (which is better anyway, expectation-wise)

zandbelt:
thanks! I'll release 1.3.2 soon

PR #60, opened 2017-05-29
access_token API: error refreshing token

gonzalad:
Make sure openidc_discover is called before
refreshing the accessToken.
Otherwise, we'll get an error since token endpoint
url will be nil.


gonzalad:
Closing: waiting more test feedback from @gonzalad

PR #63, opened 2017-05-31
allow optional claim specs when verifying JWT access tokens

bodewig:
When verifying a JWT or a Bearer token currently only the `nbf` and `exp` claims are checked as this is the default. I'd like to also verify things like `iss` and maybe `aud`. Allowing callers to provide the claim specs seemed to be the easiest approach.


bodewig:
Thanks, Hans!

zandbelt:
well, actually: thanks @bodewig; I'll add you to the AUTHORS right away!

PR #65, opened 2017-06-01
add option that forces the auth flow to be executed

bodewig:
This adds an option to forcefully re-run the code grant flow. My use case is adding additional scopes after a user has already authenticated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,7 @@ http {
                 Expiration leeway for access_token renewal.
                 If this is set, renewal will happen access_token_expires_leeway seconds before the token expiration.
                 This avoids errors in case the access_token just expires when arriving to the OAuth Resoource Server.
+             --force_reauthorize = true
           }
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -113,6 +113,7 @@ http {
                 Expiration leeway for access_token renewal.
                 If this is set, renewal will happen access_token_expires_leeway seconds before the token expiration.
                 This avoids errors in case the access_token just expires when arriving to the OAuth Resoource Server.
+             --force_reauthorize = true
```
edmorley: Should this example be `false`, given the default is `false`?

zandbelt: it's an example that is commented out; so typically when including it you'd want it to be "true"; but I agree, it is not very consistent with the other options...

bodewig: I didn't see the other examples were using the defaults, sorry. Will open a new PR to fix it.

PR #72, opened 2017-06-19
Changing function signatures. openidc_authorization_response, openidc…

nielsole:
Proposal for
https://github.com/pingidentity/lua-resty-openidc/issues/69

Changing the function signatures of two functions openidc_authorization_response and openidc.authenticate

This probably breaks a lot of existing code, so this shouldn't just be waved through. 

What do you think about this?

### DIFF
--- a/lib/resty/openidc.lua
+++ b/lib/resty/openidc.lua
@@ -157,6 +157,7 @@ local function openidc_base64_url_encode(input)
 end
 
 -- send the browser of to the OP's authorization endpoint
+-- returns nil
 local function openidc_authorize(opts, session, target_url)
   local resty_random = require "resty.random"
   local resty_string = require "resty.string"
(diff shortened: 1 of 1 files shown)


zandbelt:
`openidc_authorization_response` is a local function so it should not matter breaking its signature.

AFAICT `openidc_authenticate` doesn't change its signature really; it is just that unsuccessful handing returns the correct error parameters now; the happy path is still the same

Anything I'm missing?

lib/resty/openidc.lua
```diff
@@ -633,31 +637,37 @@ function openidc.authenticate(opts, target_url, unauth_action, session_opts)
       ngx.log(ngx.ERR, err)
       return nil, err, target_url, session
     end
-    return openidc_authorization_response(opts, session), session
+    local auth_response, err = openidc_authorization_response(opts, session)
```
nielsole: Shouldn't this just be `err = openidc_authorization_response(opts, session)`

nielsole:
I'll need to give this another thought

gene1wood:
@nielsole @zandbelt Sorry I missed this notification. If you end up changing your mind on this or re-opening, highlight me and I'll take a look.

PR #73, opened 2017-06-26
Making some session data optional

### LINKED ISSUE #69: Misleading function signature openidc_authorization_response
Hi,
I just tried to set up the project and ran into an error, that wasn't returned correctly. Below is my try at root cause analysis:
`openidc_authorization_response` returns `ngx.redirect(session.data.original_url), session`.
https://github.com/pingidentity/lua-resty-openidc/blob/master/lib/resty/openidc.lua#L363
The first `ngx.redirect` is nil and the second is the session-table. This matches the usual `nil, err` guideline and is promptly misunderstood by the caller https://github.com/pingidentity/lua-resty-openidc/blob/master/lib/resty/openidc.lua#L636 . 
Thereby the example code in the `README.md` would contain the `session` as `err` and `ngx.say` would probably fail because `session` is a table:
https://github.com/pingidentity/lua-resty-openidc

    local res, err = require("resty.openidc").authenticate(opts)

    if err then
      ngx.status = 500
      ngx.say(err)
      ngx.exit(ngx.HTTP_INTERNAL_SERVER_ERROR)
    end

https://github.com/pingidentity/lua-resty-openidc/blob/master/lib/resty/openidc.lua#L363

I have a working solution, but I am not yet sure it does the right thing in all cases and will hopefully make a pull-request eventually.

nielsole:
We had the issue that the cookie produced by this project was too big for the upstream to handle (cookie size was ~ 6kb). This lead to looking into how we can reduce the amount of data stored in the cookie.
Eventually we decided to strip out the cookie entirely before passing it on upstream (not part of this pull request). Yet we already implemented a way to make some session data optional. This helps reduce overall transmission overhead. We managed to cut cookie size in half.

The option flags that need to be set for this to be activated are:

    # The entire feature only works if this is set
    opts.session_whitelist_vars
    # Includes enc_id_token
    opts.session_enc_id_token
    # Includes session_refresh_token
    opts.session_refresh_token

id-, and access-token cannot be omitted because they are inherently needed for the proxy to function.

__New function openidc_store_session__
While implementing this feature I refactored some session accesses into openidc_store_session which writes the current values into the session. The intent is to limit the access to the global state (_session_) to as few places in the code as possible and keeping the code legible.

This also changes the return values and fixes #69 

I am open for any feedback :)

### DIFF
--- a/lib/resty/openidc.lua
+++ b/lib/resty/openidc.lua
@@ -284,6 +284,7 @@ local function openidc_access_token_expires_in(opts, expires_in)
 end
 
 -- handle a "code" authorization response from the OP
+-- returns auth_response, err, original_url
 local function openidc_authorization_response(opts, session)
   local args = ngx.req.get_uri_args()
   local err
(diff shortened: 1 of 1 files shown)


zandbelt:
Wrt. the cookie size, see also: https://github.com/pingidentity/lua-resty-openidc/issues/33

lua-resty-session should be able to chunk up the cookie by itself since release 2.15, or you could use a server-based caching mechanism

nielsole:
The chunking is working. This is more about the overall cookie size

zandbelt:
I'd rather see a single `opt` flag than the combination of two and thus blacklist the. E.g.
```
opts.session_blacklist_vars = { "enc_id_token" = true, "refresh_token" = true }
```
and then use:
```
if opts.session_blacklist_vars["enc_id_token"] then
  session.data.enc_id_token = auth_response.id_token or session.data.enc_id_token
end
```
etc.

zandbelt:
superseded by #87

PR #74, opened 2017-06-30
fix set cache when pass an invalid token to keycloak

dereknex:
I got a error like as blow:
```
 2017/06/30 08:24:01 [error] 7#7: *15 lua entry thread aborted: runtime error: /usr/local/lib/lua/5.1/resty/openidc.lua:69: bad argument #3 to 'set' (number expected, got nil)
```

It's because keycloak return `{"active":false}` when pass a invalid token.


zandbelt:
good catch, I'll merge soon

PR #78, opened 2017-07-03
Adding token hint to op logout request

pgp44:
Not sure this is ok in all cases. E.g. when no post_logout_redirect_uri  is configured. Unfortunately I don't have sufficient infrastructure to test that.


zandbelt:
If you could make this configurable with the default being "off" it would be backwards compatible.

pgp44:
That was a quick merge. I did not realise that as soon as I pushed to the fork that would become part of the pr. I am afraid it was not yet correct. I committed a fix to the fork.

PR #86, opened 2017-08-04
Use a separate flag in the session to mark it as authenticated

hanikesn:
There will be a follow up pull request to allow configuration of what's actually stored in the session.
We are dependent on keeping the session size small as we store it in a cookie for easier operations and the current sessions size is about 6kb which is prohibitive, e.g. webserver like nginx, jetty and tomcat can't deal with large headers. Also they get transfered on every request.

This allows the later removal of the id_token from the session.
Also clear nonce and state from the session


zandbelt:
included by #87

PR #87, opened 2017-08-04
Feature flags

hanikesn:
These commits introduce feature flags, which allow configuration of what features are enabled. e.g. retrieving the user info, or storing the access_key or encrypted id_token.

This includes PR #86.

We are dependent on keeping the session size small as we store it in a cookie for easier operations and the current sessions size is about 6kb which is prohibitive, e.g. webservers like nginx, jetty and tomcat can't deal with large headers. Also they get transfered on every request.

This PR also superseeds #73. Feedback is very welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,9 @@ http {
                 This avoids errors in case the access_token just expires when arriving to the OAuth Resoource Server.
              --force_reauthorize = false
              -- when force_reauthorize is set to true the authorization flow will be executed even if a token has been cached already
+             --features = {id_token=true}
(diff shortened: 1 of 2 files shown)


zandbelt:
Thanks, this looks good. I would like to nag about one thing though: I don't think "features" is a very descriptive name for this feature.

hanikesn:
Yeah I'm also not to happy with that name. How about `session_storage`?

zandbelt:
Hmm, how about `session_contents` (in both code and option name)?

hanikesn:
I renamed it to `session_contents` now.

PR #98, opened 2017-10-11
Redirect from OP tests

bodewig:
This adds tests for openidcs reactions on the authorization code response.

Next up will be tests for the token request and with that id_token validation.


bodewig:
completely part of #99

Issue #95, opened 2017-09-21
Do not try to use unsupported Auth method


betalb:
If OIDC discovery returns a list of  `token_endpoint_auth_methods_supported`, and `token_endpoint_auth_method` option was not provided, do not pick first from the list of IdP supported methods, instead pick first supported by library.

https://github.com/pingidentity/lua-resty-openidc/blob/8c15adb55b2290c80808e98b9f1503941ea87376/lib/resty/openidc.lua#L545

Out of 4 methods, described in discovery spec (`client_secret_post`, `client_secret_basic`, `client_secret_jwt`, and `private_key_jwt`) looks like only first 2 are actually supported.

Also it will be more clear if validation will be done against supplied `token_endpoint_auth_method`, that it is actually supported by library, not only that it is present in discovery response.

bodewig:
@betalb I think #96 should solve this.

PR #99, opened 2017-10-12
Token retrieval tests

bodewig:
This tests the interaction with the token endpoint and the id_token verification.

Contains #98 - also contains tests for #96 and thus #95


bodewig:
BTW, I found a few minor issues in id_token verification:

* the log message for an old `iat` is wrong - this would be correct for `nbf` which doesn't seem to be part of the OIDC spec
* when checking the `exp` claim `iat_slack` is not used while the code in `openidc.jwt_verify` uses it
* the presence of `exp` and `aud` claims should be checked explicitly rather than via Lua errors

I've got those fixed locally but don't want to create another PR depending on this one - fixing the error message means I have to modify a test.

PR #82, opened 2017-07-06
JWT Token validation using PEM key generated by n and e parameters

venkatmarepalli:
I have implemented this patch to validate tokens which doesn't have x5c entry. Please review the code. I had to hard code RSA encryption header in 64 bit format. Please feel free to make any changes.
#71

PR #104, opened 2017-10-20
Bearer token verification tests

bodewig:
This adds tests for `openidc.bearer_jwt_verify` and thus `openidc.jwt_verify`. It also contains a commented out test that should pass on your branch that contains something similar to #82 

Some things I've noticed:

* right now it is required that the header of the JWT contains a `kid` claim. This one is optional if the JWKS only contains a single key.
* at least the recent version of lua-resty-jwt I've used verifies the `exp` claim itself - and it doesn't take the `iat_slack` into account (as it's not told to). I think the check inside lua-resty-jwt  should be suppressed or it should be told to honor `iat_slack` (assertions are commented out right now)
* I'm not sure whether http://self-issued.info/docs/draft-ietf-jose-json-web-key.html#rfc.section.4.7 is the normative reference but it says `x5c` is base64 not base64url encoded. My test certificate is base64 encoded and works, so maybe the point is moot as `openidc_base64_url_decode` works for it as well
* I think it would be easy to support `x5u` as well
* in theory `x5c`, `x5u`, `n` and `e` could be claims of the JWT's header and openidc could use them. The OIDC spec discourages their usage but doesn't prohibit them either. Likely irrelevant.

I'll be happy to take a stab at either of these points. IMHO the first two should be addressed.


zandbelt:
- The normative reference for JWKs with x5c is https://tools.ietf.org/html/rfc7517#section-4.7, which makes it base64url.

- "self-contained" key info is like sending the key with the message which supposes some superseding trust model which does not exist.

Thanks for your contributions; looking forward to the next 2 indeed :-)

bodewig:
Thanks for the pointer, I should have looked for an RFC myself. It says (emphasis mine)

> Each string in the array is a base64-encoded (Section 4 of [RFC4648] -- **not base64url-encoded**) DER [ITU.X690.1994] PKIX certificate value.

given that `openidc_base64_url_decode` just works for normal base64 encoded strings as well, this doesn't matter. at all. Sorry for the nit-picking :-)

OK, I'll look into the first two, then. The `kid` part should be simple.

PR #111, opened 2017-10-27
ensure id_token_hint uses correct separator on logout

bodewig:
currently the `id_token_hint` parameter is added to `opts.redirect_after_logout_uri` separated by a `&` unconditionally.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+10/27/2017
+- ensured id_token_hint uses the correct query string separator when
+  opts.redirect_after_logout_uri is used together with redirect_after_logout_with_id_token_hint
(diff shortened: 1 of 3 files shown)


bodewig:
related, but a slightly different issue: shouldn't the `id_token_hint` also be added to the `end_session_endpoint` as well if this is the redirection target?

zandbelt:
yes, in fact I believe that would follow the standard http://openid.net/specs/openid-connect-session-1_0.html#RPLogout whereas `redirect_after_logout_uri` is somewhat undefined

bodewig:
yes, that's where I got the idea from :-)

PR follows

PR #118, opened 2017-11-10
let user re-authenticate if we know the access token has expired

### LINKED ISSUE #117: expired refresh_token is used to retrieve new access_token
#52 introduced automatic refresh of the access_token if it expires. Unfortunately, there is no check if the used refresh_token has expired when doing this. It doesn't check the "refresh_expires_in"-claim returned by the token-endpoint or the "exp"-claim of the refresh_token.

Because of this, the user only sees error responses and has no way to re-login.

Debug log:

```
2017/11/10 15:32:43 [debug] 11547#11547: *1 [lua] openidc.lua:83: openidc_cache_set(): cache set: success=true err=nil forcible=false
2017/11/10 15:32:43 [debug] 11547#11547: *1 [lua] openidc.lua:580: openidc_get_token_auth_method(): 1 => private_key_jwt
2017/11/10 15:32:43 [debug] 11547#11547: *1 [lua] openidc.lua:580: openidc_get_token_auth_method(): 2 => client_secret_basic
2017/11/10 15:32:43 [debug] 11547#11547: *1 [lua] openidc.lua:583: openidc_get_token_auth_method(): no configuration setting for option so select the first supported method specified by the OP: client_secret_basic
2017/11/10 15:32:43 [debug] 11547#11547: *1 [lua] openidc.lua:597: openidc_get_token_auth_method(): token_endpoint_auth_method result set to client_secret_basic
2017/11/10 15:32:43 [debug] 11547#11547: *1 [lua] openidc.lua:618: openidc_access_token(): refreshing expired access_token: MY_ACCESS_TOKEN with: MY_REFRESH_TOKEN
2017/11/10 15:32:43 [debug] 11547#11547: *1 [lua] openidc.lua:567: openidc_get_token_auth_method(): 1 => private_key_jwt
2017/11/10 15:32:43 [debug] 11547#11547: *1 [lua] openidc.lua:567: openidc_get_token_auth_method(): 2 => client_secret_basic
2017/11/10 15:32:43 [debug] 11547#11547: *1 [lua] openidc.lua:569: openidc_get_token_auth_method(): configured value for token_endpoint_auth_method (client_secret_basic) found in token_endpoint_auth_methods_supported in metadata
2017/11/10 15:32:43 [debug] 11547#11547: *1 [lua] openidc.lua:597: openidc_get_token_auth_method(): token_endpoint_auth_method result set to client_secret_basic
2017/11/10 15:32:43 [debug] 11547#11547: *1 [lua] openidc.lua:252: openidc_call_token_endpoint(): client_secret_basic: authorization header 'Basic MY_CREDENTIALS'
2017/11/10 15:32:43 [debug] 11547#11547: *1 [lua] openidc.lua:261: openidc_call_token_endpoint(): request body for token endpoint call: refresh_token=MY_REFRESH_TOKEN&scope=openid%20email%20profile&grant_type=refresh_token
2017/11/10 15:32:43 [debug] 11547#11547: *1 [lua] http.lua:631: send_request(): 
POST /auth/realms/ems/protocol/openid-connect/token HTTP/1.1
User-Agent: lua-resty-http/0.11 (Lua) ngx_lua/10010
Content-Length: 1221
Content-Type: application/x-www-form-urlencoded
Authorization: Basic MY_CREDENTIALS
Host: MY_AUTH_SERVER


2017/11/10 15:32:43 [debug] 11547#11547: *1 [lua] openidc.lua:276: openidc_call_token_endpoint(): token endpoint response: {"error":"invalid_grant","error_description":"Refresh token expired"}
```

I am using version 1.4.1.

bodewig:
closes #117 

With this patch if we know the token is expired and we fail to refresh it the user will be sent to the authorization endpoint.

While looking through the code I realized that if we know the user has authenticated once (`session.data.authenticated` is true) but don't store the access token because of `store_in_session` settings then all requests will be allowed as long as the session exists. I'm not sure this is intended.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+11/13/2017
+- provided explicit control over whether the access token gets renewed
+  on expiry. If refresh fails, the user is redirected to the OP's
+  authorization endpoint.
+
 11/06/2017
 - added support for configurable network timeouts
 
--- a/README.md
+++ b/README.md
@@ -108,9 +108,12 @@ http {
              --redirect_after_logout_with_id_token_hint = true,
              --token_endpoint_auth_method = ["client_secret_basic"|"client_secret_post"],
              --ssl_verify = "no"
+
+             --renew_access_token_on_expiry = true
+             -- whether this plugin shall try to silently renew the access token once it is expired if a refresh token is available.
+             -- if it fails to renew the token, the user will be redirected to the authorization endpoint.
              --access_token_expires_in = 3600
              -- Default lifetime in seconds of the access_token if no expires_in attribute is present in the token endpoint response. 
-             -- This plugin will silently renew the access_token once it is expired if refreshToken scope is present.
 
              --access_token_expires_leeway = 0
              --  Expiration leeway for access_token renewal. If this is set, renewal will happen access_token_expires_leeway seconds before the token expiration. This avoids errors in case the access_token just expires when arriving to the OAuth Resource Server.
--- a/lib/resty/openidc.lua
+++ b/lib/resty/openidc.lua
@@ -953,7 +953,10 @@ function openidc.authenticate(opts, target_url, unauth_action, session_opts)
   end
 
   local token_expired = false
-  if session.present and session.data.authenticated and store_in_session(opts, 'access_token') then
+  local try_to_renew = opts.renew_access_token_on_expiry == nil or opts.renew_access_token_on_expiry
+  if try_to_renew and session.present and session.data.authenticated
+    and store_in_session(opts, 'access_token')
+  then
     -- refresh access_token if necessary
     access_token, err = openidc_access_token(opts, session)
     if err then
@@ -970,7 +973,7 @@ function openidc.authenticate(opts, target_url, unauth_action, session_opts)
   if not session.present
     or not (session.data.id_token or session.data.authenticated)
     or opts.force_reauthorize
-    or token_expired
+    or (try_to_renew and token_expired)
   then
     if unauth_action == "pass" then
       return
--- a/lib/resty/openidc.lua
+++ b/lib/resty/openidc.lua
@@ -875,7 +875,7 @@ local function openidc_access_token(opts, session)
     return session.data.access_token, err
   end
   if session.data.refresh_token == nil then
-    return nil, err
+    return nil, "token expired and no refresh token available"
   end
 
   ngx.log(ngx.DEBUG, "refreshing expired access_token: ", session.data.access_token, " with: ", session.data.refresh_token)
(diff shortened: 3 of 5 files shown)


zandbelt:
If that matches your PR you can merge!

bodewig:
Well, I think openidc seems to be used in the role of a resource server as well (not only by myself :-). The `access_token` and `introspect` methods go beyond RP. But I really don't want to push things further if you feel this is going into the wrong direction.

This PR doesn't change the behavior of 1.4.1 in the following situations:

* the access token is never stored in the session at all
* the access token is stored in the session and hasn't expired
* the access token is stored in the session along with a refresh token, the access token has expired and we manage to refresh it

The PR does change the behavior of 1.4.1 in the following ways:

* the access token has been stored in the session, has expired and there is no refresh token. In 1.4.1 `authenticate` would return successful (with a `nil` `access_token` and a `nil` `err`) where it now redirects to the authorization endpoint
* the access token has been stored in the session, has expired and an error occurs when refreshing it. In 1.4.1 authenticate would return with a `nil` `access_token` and `err` set to a non-`nil`l value. Now the error will be logged and the request gets redirected to the authorization endpoint (which seems to be what #117 asks for).

If this is OK with you, I'll be happy to merge (or let you do so).

An alternative way to fix #117 would be for the calling code to somehow detect this specific kind of error occurred, set session.data.authenticated to false and re-call authenticate.

zandbelt:
When lua-resty-openidc is used in the Resource Server role, one would use `bearer_jwt_verify` or `introspect` which does not create a session and no authorization endpoint is in play.

bodewig:
I'm not sure it is always either RP or Resource Server. I defer to you to decide whether this change is acceptable or not.

zandbelt:
How about making this behavior ("keep access token fresh when running as OIDC RP") configurable (and I'd be ok with the default being "true") so that we have backwards compatibility? I'm a bit worried about an unintended side effect on a "regular" RP that does the following: compete the authentication flow and use the access token once to obtain userinfo, then forget about it and just maintain a local RP session.

bodewig:
Understood. I'll improve this PR along these lines. My track record of naming options isn't too good, though :-) `opts.ensure_fresh_access_token`?

ChristianCiach:
> An alternative way to fix #117 would be for the calling code to somehow detect this specific kind of error occurred, set session.data.authenticated to false and re-call authenticate.

I think this would be very prone for error. I didn't even know that I can access the session in this way.

ChristianCiach:
Btw, we are currently using mod_auth_openidc and I am trying to migrate to OpenResty. As far as I can see, mod_auth_openidc always redirects to the auth-page when the access_token expires. I expected lua-resty-openidc to behave in the same way.

ChristianCiach:
> As far as I can see, mod_auth_openidc always redirects to the auth-page when the access_token expires.

I am probably wrong about this. The difference seems to be that mod_auth_openidc uses "OIDCSessionMaxDuration" which defaults to 8 hours. It seems like mod_auth_openidc doesn't automatically refreshes the access_token, so there is no error shown to the user even if is expires.

zandbelt:
`opts.ensure_fresh_access_token` or `opts.renew_access_token_on_expiry` would be good IMHO

bodewig:
added `opts.renew_access_token_on_expiry` and a test for the case where it is set to `false` - in which case `authenticate` doesn't even try to refresh the access token.

PR #119, opened 2017-11-10
Since we force `opts.prompt = "none"` below the option is retained.

gdestuynder:
This means if the code is re-called without re-loading `opts` the call
to authenticate will be done with `opts.prompt = "none"` again, where
user interaction would be expected.
This happens if `opts` is declared as a global value in openresty/nginx,
such as by:
`init_by_lua_file "/usr/local/openresty/nginx/conf/conf.d/server.lua";`
in `nginx.conf`, where `server.lua` contains the `opts` structure.
This happens when the server owner needs a global `opts` declaration
reachable by all `server {}` structures in nginx.

### DIFF
--- a/lib/resty/openidc.lua
+++ b/lib/resty/openidc.lua
@@ -962,6 +962,7 @@ function openidc.authenticate(opts, target_url, unauth_action, session_opts)
         target_url,
         session
     end
+    opts.prompt = nil
     openidc_authorize(opts, session, target_url)
     return nil, nil, target_url, session
   end
--- a/tests/spec/redirect_to_op_spec.lua
+++ b/tests/spec/redirect_to_op_spec.lua
@@ -49,18 +49,6 @@ describe("when accessing the custom protected resource without token", function(
   end)
 end)
 
-describe("when explicitly asking for a prompt parameter", function()
-  test_support.start_server({oidc_opts = {prompt = "none"}})
-  teardown(test_support.stop_server)
-  local _, status, headers = http.request({
-    url = "http://127.0.0.1/default/t",
-    redirect = false
-  })
-  it("then it is included", function()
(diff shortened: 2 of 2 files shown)


gdestuynder:
Note: I had to drop the test checking that the `prompt` parameter set in `opts` is honoured - because we did not in fact honour it in some conditions (the test would not see that), and with this fix, we never honour it.
If we do, it would break functionality - and since `opts` is not only used as a user parameter but also as a structure to pass data internally, I didn't find another solution that wouldn't require either refactoring a lot of code, or having a function that behaves differently than other functions :(

zandbelt:
I guess a better approach would be to explicitly pass a prompt parameter into `authenticate` and `openidc_authorize` though that changes the function signature and backwards compatibility...?

bodewig:
with this PR we'd break code that explicitly asks for the prompt parameter when calling `authenticate` (which is what the test checks ;-). I'd prefer an explicit parameter over silently discarding the options set.

One option not breaking backwards compatibility could be to only add the explicit parameter to `openidc_authorize`, set it with `opts.prompt` in most cases and set it to `none` in the "silently reauthenticate" case without changing `opts` there at all. That way `opts` wouldn't get modified by `authenticate`. Would this work for you @gdestuynder ?

bodewig:
@gdestuynder https://github.com/bodewig/lua-resty-openidc/commit/195c9555f37a06ead92f2f60a08fa51606cc93e6 is an alternative solution

zandbelt:
I'd prefer Stefan's approach if acceptable @gdestuynder

gdestuynder:
@ bodewig that sounds alright

bodewig:
thanks, see #120

gdestuynder:
hah great thank you!

PR #129, opened 2017-12-22
use X-Forwarded-Host when building redir uri if present

### LINKED ISSUE #128: Support of X-Forwarded-Host
Hi all,
In case both lua-resty-openidc (deployed as Relying Party) and the Authorization server are deployed behind a front-end reverse proxy, hostnames of URLs used "externally" (i.e. from the Web Browser to the front-end reverse proxy) are different from the hostnames of URLs used "internally (i.e. from the front-end reverse proxy to the Relying Party and the Authorization server).
As a consequence, when lua-resty-openidc is called, internal hostnames are used, so the way the redirect URI is build ([here](https://github.com/zmartzone/lua-resty-openidc/blob/master/lib/resty/openidc.lua#L175), using ` ngx.var.http_host`) makes the `redirect_uri` query parameter of the returned URL contain an "internal" hostname that will unfortunately not be accessible from the external Web Browser.

To prevent from this situation, the `X-Forwarded-Host` header is usually used. This header will be set by the front-end reverse proxy.
Impact on the code of lua-resty-openidc is quite limited in a first approach. Instead of:
```
return scheme.."://"..ngx.var.http_host ..opts.redirect_uri_path
```

we could have:
```
local host = ngx.req.get_headers()['X-Forwarded-Host'] or ngx.var.http_host
return scheme.."://"..host ..opts.redirect_uri_path
```

In a second approach :P, it looks the `X-Forwarded-Host` header might be multi-valued, each value being separated by a comma `","` (for example: ` X-Forwarded-Host: host1, host2`). A better implementation would be to parse the content of `X-Forwarded-Host` and use the first host of the list… 
```
local host = ngx.var.http_host
local forwarded_host = ngx.req.get_headers()['X-Forwarded-Host']
if forwarded_host then
    local sep = ","
    local hosts = string.gmatch (forwarded_host, "([^"..sep.."]+)")
    host = hosts()
end
return scheme.."://"..host ..opts.redirect_uri_path
```

It is maybe required to remove some leading or trailing spaces...

bodewig:
closes #128

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,6 @@
+12/30/2017
+- use X-Forwarded-Host header when building the redirect URI if present. See #128
+
 12/14/2017
 - release 1.5.2
 
--- a/lib/resty/openidc.lua
+++ b/lib/resty/openidc.lua
@@ -165,12 +165,25 @@ local function openidc_validate_id_token(opts, id_token, nonce)
   return true
 end
 
+local function get_first_header_and_strip_whitespace(header_name)
+  local header = ngx.req.get_headers()[header_name]
+  if header and type(header) == 'table' then
+    header = header[1]
+  end
+  return header and header:gsub('%s', '')
+end
+
 local function get_scheme()
-  return ngx.req.get_headers()['X-Forwarded-Proto'] or ngx.var.scheme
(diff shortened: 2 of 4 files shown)


aidansteele:
@zandbelt I'm pretty excited to be able to use this functionality, approximately when should we expect a release?

Issue #45, opened 2017-03-29
outgoing proxy support


ConanMalone:
Hi, we've got lua-resty-openidc working, and are now trying to get it to work in an environment that requires all traffic to go through an outbound proxy.  We are getting a connection refused error when trying to access the discovery URL.  We have installed the lua-resty-socket module but we're amateurs at lua.  Could you point us in the right direction to get openidc to use a proxy?

zandbelt:
I would have liked to point you to the support for that in the underlying lua-resty-http module that lua-resty-openidc uses to make such calls. But apparently that support does not exist there, see: https://github.com/pintsized/lua-resty-http/issues/63

So it would require modifications to lua-resty-openidc (or preferably lua-resty-http...) to make this work. Making this an enhancement request

zandbelt:
there's work-in-progress here: https://github.com/sjakthol/lua-resty-http/commit/0939833b0cca2a50054bf1db437db62295c1e299

grrolland:
Hello,

I create a pull request to fix this issue. Let me know if this could be merged with the master branche.

Regards.

zandbelt:
Nice work! Does it require a particular version of lua-resty-http?

grrolland:
Yes, this patch require the version 0.11 of lua-resty-http.

PR #139, opened 2018-01-29
Fix zmartzone/lua-resty-openidc#45 : Add outgoing proxy support

grrolland:
Hello,

This is a pull request to fix the zmartzone/lua-resty-openidc#45 issue. I'm a lua beginner so my implementation could be naïve. It's work with my use case so I propose my fix to the project : 

* Naïve implementation works with OpendId Connect Authorization
  Code Flow

* Add proxy_opts table to the opts, with two values :
   - http_proxy (in curl format : http://<proxy_host>:<proxy_port>/)
   - https_proxy (in curl format : http://<proxy_host>:<proxy_port>/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,13 @@ http {
              -- You can specify timeouts for connect/send/read as a single number (setting all timeouts) or as a table. Values are in milliseconds
              -- timeout = 1000
              -- timeout = { connect = 500, send = 1000, read = 1000 }
+             
+             -- Optionnal : use outgoing proxy to the OpenID Connect provider endpoints with the proxy_opts table : 
+             -- proxy_opts = {
+             --    http_proxy  = "http://<proxy_host>:<proxy_port>/",
+             --    https_proxy = "http://<proxy_host>:<proxy_port>/"
+             -- }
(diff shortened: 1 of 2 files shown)


zandbelt:
looks good; could you add something to the README.md sample config that shows its usage?

bodewig:
Looks good to me with a nit. You are checking for `opts.proxy_opts` several times and sometimes you also check it is a table and sometimes you don't. I'd prefer it to be consistent.

grrolland:
Hello
I rewrite the patch to be more coherent. 
Add the sample config in the README.md.

bodewig:
Many thanks!

bodewig:
I think you also need to add the proxy cinfiguration at https://github.com/grrolland/lua-resty-openidc/blob/1e81796021dcf9f06d6ff9aa42e21eac5d935a7d/lib/resty/openidc.lua#L396

grrolland:
You are right. My test case don't use the userinfo endpoint, so I don't see this function. I fix it.

bodewig:
looks great, thanks

PR #140, opened 2018-02-05
Adding usefull logs for debug or during configuration setup

pamiel:
Let’s keep it simple for my first contribution to this project :P

I’ve deployed `lua-resty-openidc` a couple of time now, and I realized that I was lacking some key logs to understand why the current configuration I setup was not working properly. Indeed, there are many logs related to the discovery endpoint or to the authentication method on the token endpoint… but nothing really useful to understand where we are in the overall OpenID Connect process.

This (simple!) PR consequently intends to add these logs to trace the key steps of the authentication process, for a better debugging to make the deployment configuration work.

### DIFF
--- a/lib/resty/openidc.lua
+++ b/lib/resty/openidc.lua
@@ -800,6 +800,8 @@ local function openidc_authorization_response(opts, session)
     state = session.data.state
   }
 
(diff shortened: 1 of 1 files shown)


bodewig:
I can see how this may be useful, many thanks!

PR #142, opened 2018-02-07
Adding functions to invalidate the various caches

pamiel:
A time to live in the server caches for the data retrieved from Discovery and JWKS endpoints is defined (1day). This is good as it enables implementing some specific use cases such as key rotation.

However, it might be required to invalidate the caches immediately, in order an update of the OP to be taken into account immediately, for example.

This PR provide the functions (one internal, and one external) to enable a `lua-resty-openidc` client code to trigger the caches invalidation. Discovery and JWKS, caches are flushed, but also introspection as it is better to re-evaluate tokens as the jwks and discovery data may have changed.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,6 @@
+02/08/2018
+- added functions to manually request the invalidation of the caches; see #142
+
 02/06/2018
 - added more logging that may help debugging configuration; see #140; thanks @pamiel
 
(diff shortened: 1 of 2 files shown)


zandbelt:
Could you please add something to the ChangeLog as well? Thanks.

pamiel:
Sure. Pushed.

pamiel:
Hi,
Any missing info and blocking point that prevents from accepting the PR ?

bodewig:
no, everything is fine, sorry for the delay

pamiel:
No problem; I thought I forgot to do something !
Thanks a lot.

PR #145, opened 2018-02-19
Update doc to reflect JWKS validation behavior when x5c isn't present but module and exponent are

arcivanov:
Update doc to reflect JWKS validation behavior when x5c isn't present but module and exponent are


bodewig:
many thanks!

bodewig:
Merged and tweaked the sentence myself, thanks again.

PR #154, opened 2018-04-20
- added support for passing bearer token as cookie

dmitriyblok:
- added support for introspection interval
- added support for passing request cookies on introspection call
- added passing introspection endpoint authorization method into openidc_call_token_endpoint call

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+04/19/2018
+- added support for passing bearer token as cookie
+- added support introspection interval
+- added support for passing request cookies on introspection call
+- added passing introspection endpoint authorization method into openidc_call_token_endpoint call 
+
 04/11/2018
 - bugfix rare case where OP public key is starting with 0x80 byte, and thus resulting in
   systematic token signature verification error with message "too long"
--- a/README.md
+++ b/README.md
@@ -325,6 +325,93 @@ http {
   }
 }
 ```
+## Sample Configuration for passing bearer OAuth 2.0 access tokens as cookie
+
+Sample `nginx.conf` configuration for validating Bearer Access Tokens passed as cookie against a ORY/Hydra Authorization Server.
+
+```
+events {
+  worker_connections 128;
+}
+
+http {
+
+  lua_package_path '~/lua/?.lua;;';
+
+  resolver 8.8.8.8;
+
+  lua_ssl_trusted_certificate /opt/local/etc/openssl/cert.pem;
+  lua_ssl_verify_depth 5;
+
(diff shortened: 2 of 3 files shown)


zandbelt:
lgtm

bodewig:
Sorry, I'll need more time for a real review.

I think you can access cookie values directly via `ngx.var['cookie_'..COOKIE_NAME]` without us having to parse the header(s) ourselves. At least this is how I've done it myself in the past, but maybe there is a good reason to avoid it.

dmitriyblok:
From what I understand parsing headers is more universal and will work in all cases whereas $cookie_COOKIE_NAME only works with alphanumeric cookies. https://stackoverflow.com/questions/26128412/how-to-extract-some-value-from-cookie-in-nginx

bodewig:
I knew I'd lear something by asking, thanks :-)

The table indexing syntax works for me for cookie names containing dashes and lua-resty-session also uses `ngx.var['cookie_X']` for cookies containing underscores in their name (see https://github.com/bungle/lua-resty-session/blob/master/lib/resty/session.lua#L143), so I guess this syntax is safe as opposed to `ngx.var.cookie_X`.

dmitriyblok:
OK. I can change it to use ngx.var["cookie_X"]

bodewig:
I wasn't aware of the existing options of `mod_auth_openidc` as I have never used that myself. Hans will probably have a better idea about names then :-)

For tests I really just run the two docker command from the bottom of the readme. The developer experience isn't nice, of course. The roundtrip of creating new docker image with every change is ugly and often I end up editing the test cases inside of the container until the tests pass. No better idea, I'm sorry.

zandbelt:
well the net result of removing the introspection results from the cache is that the token will be introspected again, hence the repetition interval

bodewig:
OK, fine with me. We can merge this now or wait for the REAME updates.

dmitriyblok:
I have updated README with the new example and descriptions shamelessly stolen from mod_oauth_openidc :-)

bodewig:
Many thanks!

I'm not sure "(can be configured on a per-path basis)" is quite true - of course you can use different opts for different locations, but that's true for all settings.

dmitriyblok:
Ugh-ogh. Bad copy paste. Could you remove "(can be configured on a per-path basis)" from there? Thank you!

bodewig:
tests added with eb8510c and abc780a and addressed some of my nits with e3ce0d1 and 5a9ac8e

It would be good if you could have a second look @dmitriyblok

Issue #157, opened 2018-05-17
Invalid ID Token after refreshing tokens with the token endpoint


grrolland:
After refreshing automaticaly tokens (Access en Refresh) with the token endpoind, the ID token in session is not updated with a new id token from the providers. 

This let the session in an incoherent state with the other tokens store in it.

grrolland:
Just added a pull request to fix this.

grrolland:
Done all test passed.

grrolland:
Add compatibility with session_contents opts in the pull request.

PR #158, opened 2018-05-17
Fix zmartzone/lua-resty-openidc#157 : Update Id token in session afte…

grrolland:
This pull request update the id token in session after refreshing tokens with token endpoint. Finally, the content of the session is coherent (access token, refresh token, id token).

### DIFF
--- a/lib/resty/openidc.lua
+++ b/lib/resty/openidc.lua
@@ -829,6 +829,44 @@ local function openidc_load_jwt_and_verify_crypto(opts, jwt_string, asymmetric_s
   return jwt_obj
 end
 
+--
+-- Load and validate id token from the id_token properties of the token endpoint response
+-- Parameters :
+--     - opts the openidc module options
+--     - jwt_id_token the id_token from the id_token properties of the token endpoint response
+--     - session the current session
+-- Return the id_token, nil if valid
+-- Return nil, the error if invalid
+--
+local function openidc_load_and_validate_jwt_id_token(opts, jwt_id_token, session)
+
+  local jwt_obj, err = openidc_load_jwt_and_verify_crypto(opts, jwt_id_token, opts.secret, opts.client_secret)
+  if err then
+    local is_unsupported_signature_error = jwt_obj and not jwt_obj.verified and not is_algorithm_supported(jwt_obj.header)
+    if is_unsupported_signature_error then
+      ngx.log(ngx.WARN, "ignored id_token signature as algorithm '" .. jwt_obj.header.alg .. "' is not supported")
+    else
+      local alg = (jwt_obj and jwt_obj.header and jwt_obj.header.alg) or ''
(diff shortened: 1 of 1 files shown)


bodewig:
I agree, in that case the PR looks good to me.

zandbelt:
It would be great to include a test for it... (never thought I would say that ;-))

bodewig:
I've pushed the PR, a small refactoring (please have a look at 11a9f26 @grrolland ) and tests to the pr-158 branch.

While thinking through the case where the token endpoint returns an invalid id token on refresh I feel the handling is inconsistent right now. `openidc_access_token` (correctly IMHO) returns an error in that case but we keep the access token stored inside the session. It seems we should remove all tokens from the session if we want to discard the token endpoint's response.

zandbelt:
I don't see that last implication: whatever access token is returned from the token endpoint may be stored in the session; even if there's an error in a subsequent call to the token endpoint, we can still choose to preserve the old one under the assumption "if it still valid it is ok otherwise we'll find out sooner or later"

bodewig:
As the current code stands the old access token has been replaced by the one that was returned from the token endpoint inside of the session. The inconsistency I see is that we don't return the access token because id token verification failed but we still store it inside the session. I'd be fine with either removing the access token completely or keeping the old one, but the current state feels wrong to me.

zandbelt:
Ah right, I misunderstood at first but agree with you.

bodewig:
What would be your preferred option, remove clear the tokens completely or restore the old ones?

I don't think the old ones will be useful as the refresh token has already been used by now and the access token has likely expired, so would opt for clearing the tokens.

zandbelt:
Yeah, you're right: the reason for refreshing should be that the client can't use the existing access token so there should be no reason to keep it regardless of the result. That is different from the id_token.

bodewig:
OK, I've changed the branch, see a968e76

zandbelt:
So now the `id_token` is also set to `nil` where it doesn't  need to https://github.com/zmartzone/lua-resty-openidc/pull/158#issuecomment-389979084

bodewig:
Ah, thanks, you are right. 551564f

zandbelt:
the log message "ngx.log(ngx.ERR, "invalid id token, discaring refreshed id token")
" is off (+ a spelling error)

bodewig:
Not sure I know what you mean with "off", see b27009f

Unfortunately I've got some second thoughts again, sorry for that. Wouldn't it be cleaner if we verified an id_token returned while refreshing regardless of whether we are going to store it inside the session? In that case we could treat it like a "normal" token refresh error and simply leave the whole session alone (see https://github.com/zmartzone/lua-resty-openidc/blob/master/lib/resty/openidc.lua#L1090 ).

bodewig:
regarding those latest thoughts see 13516c4 in yet another branch that I'll need to clean up once we are happy :-)

bodewig:
merged, thanks @grrolland !

PR #165, opened 2018-06-06
Update lua-resty-jwt for OpenSSL 1.1

cdbattags:
This should add support for OpenSSL 1.1 with the new fork of https://github.com/SkyLothar/lua-resty-jwt at https://github.com/cdbattags/lua-resty-jwt!

Still very confused that you're using two different versions of `lua-resty-hmac` with OPM version and LuaRocks...

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -4,4 +4,4 @@ author = Hans Zandbelt (@zandbelt)
 is_original = yes
 license = apache2
 repo_link = https://github.com/zmartzone/lua-resty-openidc
-requires = openresty, pintsized/lua-resty-http >= 0.08, bungle/lua-resty-session >= 2.8, cdbattags/lua-resty-jwt >= 0.2.0, jkeys089/lua-resty-hmac >= 0.2
+requires = openresty, pintsized/lua-resty-http >= 0.08, bungle/lua-resty-session >= 2.8, cdbattags/lua-resty-jwt >= 0.2.0
--- a/dist.ini
+++ b/dist.ini
@@ -4,4 +4,4 @@ author = Hans Zandbelt (@zandbelt)
 is_original = yes
 license = apache2
 repo_link = https://github.com/zmartzone/lua-resty-openidc
-requires = openresty, pintsized/lua-resty-http >= 0.08, bungle/lua-resty-session >= 2.8, SkyLothar/lua-resty-jwt >= 0.1.5, jkeys089/lua-resty-hmac
+requires = openresty, pintsized/lua-resty-http >= 0.08, bungle/lua-resty-session >= 2.8, cdbattags/lua-resty-jwt >= 0.2.0, jkeys089/lua-resty-hmac >= 0.2
--- a/lua-resty-openidc-1.5.4-1.rockspec
+++ b/lua-resty-openidc-1.5.4-1.rockspec
@@ -25,7 +25,7 @@ dependencies = {
     "lua >= 5.1",
     "lua-resty-http >= 0.08",
     "lua-resty-session >= 2.8",
-    "lua-resty-jwt >= 0.1.5",
+    "lua-resty-jwt >= 0.2.0",
     "lua-resty-hmac"
 }
 build = {
--- a/lua-resty-openidc-1.5.4-1.rockspec
+++ b/lua-resty-openidc-1.5.4-1.rockspec
@@ -25,8 +25,7 @@ dependencies = {
     "lua >= 5.1",
     "lua-resty-http >= 0.08",
     "lua-resty-session >= 2.8",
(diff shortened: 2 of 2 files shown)


ChristianCiach:
I highly recommend to merge this, because the current dependency setup breaks new and updated installations of OpenResty with lua-resty-openidc

ChristianCiach:
This is offtopic, but a good moment to ask about your opinion: 

I know that OpenResty itself strongly discourages the use of LuaRocks in favor of OPM. What do you think about that? I have the feeling that LuaRocks is a lot more mature than OPM. And in this specific case, using LuaRocks would have fixed this issue a lot faster, because it would have pulled your fork of lua-resty-jwt.

cdbattags:
OPM is extremely early on to be quite honest. The fact that I couldn't even delete from the registry is crazy to me. Only problem with this though is that `lua-resty-hmac` that's used in the OPM package is completely different than the one in the LuaRocks package... which is concerning.

So if we can get https://github.com/jkeys089/lua-resty-hmac on LuaRocks then we'd be set. Because in the current state on LuaRocks I still don't think this is fully OpenSSL 1.1 ready without the changes @jkeys089 has in his package.

dist.ini
```diff
@@ -4,4 +4,4 @@ author = Hans Zandbelt (@zandbelt)
 is_original = yes
 license = apache2
 repo_link = https://github.com/zmartzone/lua-resty-openidc
-requires = openresty, pintsized/lua-resty-http >= 0.08, bungle/lua-resty-session >= 2.8, SkyLothar/lua-resty-jwt >= 0.1.5, jkeys089/lua-resty-hmac
+requires = openresty, pintsized/lua-resty-http >= 0.08, bungle/lua-resty-session >= 2.8, cdbattags/lua-resty-jwt >= 0.2.0, jkeys089/lua-resty-hmac >= 0.2
```
zandbelt: so jkeys089/lua-resty-hmac >= 0.2 now but I only see 0.0.2 in the OPM repo

lua-resty-openidc-1.5.4-1.rockspec
```diff
@@ -25,7 +25,7 @@ dependencies = {
     "lua >= 5.1",
     "lua-resty-http >= 0.08",
     "lua-resty-session >= 2.8",
-    "lua-resty-jwt >= 0.1.5",
+    "lua-resty-jwt >= 0.2.0",
     "lua-resty-hmac"
```
zandbelt: I guess because lua-resty-jwt comes with its own (/vendor) version of lua-resty-hmac, this dependency can be removed (Travis CI would verify this)

cdbattags: My thoughts exactly.

dist.ini
```diff
@@ -4,4 +4,4 @@ author = Hans Zandbelt (@zandbelt)
 is_original = yes
 license = apache2
 repo_link = https://github.com/zmartzone/lua-resty-openidc
-requires = openresty, pintsized/lua-resty-http >= 0.08, bungle/lua-resty-session >= 2.8, cdbattags/lua-resty-jwt >= 0.2.0, jkeys089/lua-resty-hmac >= 0.2
+requires = openresty, pintsized/lua-resty-http >= 0.08, bungle/lua-resty-session >= 2.8, cdbattags/lua-resty-jwt >= 0.2.0
```
zandbelt: ok, I did not mean to remove it from the OPM dist.ini: AFAIK that format doesn't have the option to explicitly include dependencies

cdbattags:
Sorry, this double package manager stuff is quite confusing... =P

zandbelt:
Ok, I'm confused too :-)
I guess you we're right about removing the explicit `lua-resty-hmac` dependency from `lua-resty-openidc` OPM dist.ini file anyhow: since `lua-resty-jwt` takes care of its own dependencies (and you've specified this dependency in the dist.ini of your `lua-resty-jwt`.

But: that brings us back to this question https://github.com/zmartzone/lua-resty-openidc/pull/165#issue-193114790 since it really is `lua-resty-jwt` that includes its own version of `lua-resty-hmac` in luarocks, but pulls `jkeys089/lua-resty-hmac` from the repo in OPM ...

cdbattags:
So I guess let us worry about the dep then? Since you never explicitly call it anywhere in yours that makes the most sense, no?

zandbelt:
sounds good

cdbattags:
Ok, updating this one last time then without dist.ini dep and we wait for Travis and then go for it.

cdbattags:
Ok, should be good to go! Can we get this merged and fresh uploads to OPM and LuaRocks?

PR #171, opened 2018-07-04
add customized expiration time option for discovery and jwk cache

dhlin:


zandbelt:
Looks good to me; @bodewig  Whaddayasay? Should we merge this and release 1.6.1 together with the key length fix?

bodewig:
yep,, +1 to both.

Thanks @dhlin

PR #172, opened 2018-07-07
Add support for Windows Integrated Authentication with AD FS 4.0 or newer

vishnu2kmohan:
`client_secret` must not be set if Windows Integrated Authentication (WIA) is used with Active Directory Federation Services (AD FS) 4.0 (or newer) on Windows Server 2016 (or newer)


vishnu2kmohan:
I (too) didn't add a test for this PR because I picked up just enough Lua to understand what I needed to do to add Windows Integrated Authentication for my use-case: https://github.com/dcos-labs/dcos-jupyter-service/blob/v1.11.3-0.32.1/nginx/lua/oidc_auth_proxy_pass.lua

bodewig:
Is there any doc that states how to use AD FS with WIA? Does AD FS really support both client_secret_basic and client_secret_post?

vishnu2kmohan:
There's only this doc that I could find that describes that `client_secret` is optional: 
```
Note: The client's credential does not need to be a client_secret. AD FS supports the ability to use certificates or Windows Integrated Authentication as well.
```
Ref: https://docs.microsoft.com/en-us/windows-server/identity/ad-fs/overview/ad-fs-scenarios-for-developers#oauth-confidential-client

w.r.t `client_secret_post`, it's supported by Azure AD: https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-protocols-openid-connect-code#openid-connect-metadata-document

bodewig:
Thanks. I asked about the auth method as this PR changes the code for both `client_secret_post` and `client_secret_basic`, if ADFS only supports one of them, then I'd prefer to restrict the change to that.

We also include the client secret when calling the token introspection endpoint (see `openidc.introspect`). Not sure ADFS would support that, though.

vishnu2kmohan:
AD FS supports both methods, here's the https://adfs.example.com/adfs/.well-known/openid-configuration from the instance I've been working with:

```
{
    "access_token_issuer": "http://adfs.example.com/adfs/services/trust",
    "as_access_token_token_binding_supported": true,
    "as_refresh_token_token_binding_supported": true,
    "authorization_endpoint": "https://adfs.example.com/adfs/oauth2/authorize/",
    "capabilities": [],
    "claims_supported": [
        "aud",
        "iss",
        "iat",
        "exp",
        "auth_time",
        "nonce",
        "at_hash",
        "c_hash",
        "sub",
        "upn",
        "unique_name",
        "pwd_url",
        "pwd_exp",
        "mfa_auth_time",
        "sid"
    ],
    "end_session_endpoint": "https://adfs.example.com/adfs/oauth2/logout",
    "frontchannel_logout_session_supported": true,
    "frontchannel_logout_supported": true,
    "grant_types_supported": [
        "authorization_code",
        "refresh_token",
        "client_credentials",
        "urn:ietf:params:oauth:grant-type:jwt-bearer",
        "implicit",
        "password",
        "srv_challenge"
    ],
    "id_token_signing_alg_values_supported": [
        "RS256"
    ],
    "issuer": "https://adfs.example.com/adfs",
    "jwks_uri": "https://adfs.example.com/adfs/discovery/keys",
    "microsoft_multi_refresh_token": true,
    "op_id_token_token_binding_supported": true,
    "resource_access_token_token_binding_supported": true,
    "response_modes_supported": [
        "query",
        "fragment",
        "form_post"
    ],
    "response_types_supported": [
        "code",
        "id_token",
        "code id_token",
        "id_token token",
        "code token",
        "code id_token token"
    ],
    "rp_id_token_token_binding_supported": true,
    "scopes_supported": [
        "winhello_cert",
        "vpn_cert",
        "user_impersonation",
        "aza",
        "allatclaims",
        "email",
        "logon_cert",
        "openid",
        "profile"
    ],
    "subject_types_supported": [
        "pairwise"
    ],
    "token_endpoint": "https://adfs.example.com/adfs/oauth2/token/",
    "token_endpoint_auth_methods_supported": [
        "client_secret_post",
        "client_secret_basic",
        "private_key_jwt",
        "windows_client_authentication"
    ],
    "token_endpoint_auth_signing_alg_values_supported": [
        "RS256"
    ],
    "userinfo_endpoint": "https://adfs.example.com/adfs/userinfo"
}
```

zandbelt:
This PR essentially adds support for public cilents, not authenticating to the Authorization Server. Since lua-resty-openidc is a web client that can perfectly store secrets in a secure way, there shouldn't be a need to configure it as a - less secure - public client when it can be a confidential client.

Why wouldn't you use client_secret_basic or client_secret_post? I don't think ADFS would prohibit that.

vishnu2kmohan:
> I don't think ADFS would prohibit that.
But it doesn't work if you use/set either `client_secret_basic` or `client_secret_post`, it's the MSFT way or the highway :man_shrugging: 

I've tried and the Windows guys said we shouldn't be setting a secret when using Windows Integrated Authentication because the browser negotiates credentials using SSPI/GSSAPI Negotiation directly.

zandbelt:
Sorry for the delay; I guess there's value in adding support for public clients and ADFS 4.0. If you could add an entry to the ChangeLog we'll merge.

vishnu2kmohan:
Thank you for your consideration. I've updated the ChangeLog.

zandbelt:
also thanks for the date/year corrections!

PR #192, opened 2018-09-05
bypass introspection cache on demand.

dmitriyblok:
Added ability to bypass introspection cache. I found that I needed in certain situations for calls to always introspect the token regardless of whether it is cached or not and not affect the state of cache.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -28,4 +28,4 @@ reporting bugs, providing fixes, suggesting useful features or other:
 	Thomas Leplus <https://github.com/thomasleplus>
 	Donghang Lin <https://github.com/dhlin>
 	Arcadiy Ivanov <https://github.com/arcivanov>
-        Dmitriy Blok <https://github.com/dmitriyblok>
+	Dmitriy Blok <https://github.com/dmitriyblok>
--- a/AUTHORS
+++ b/AUTHORS
@@ -28,3 +28,4 @@ reporting bugs, providing fixes, suggesting useful features or other:
 	Thomas Leplus <https://github.com/thomasleplus>
 	Donghang Lin <https://github.com/dhlin>
 	Arcadiy Ivanov <https://github.com/arcivanov>
+        Dmitriy Blok <https://github.com/dmitriyblok>
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,7 +1,7 @@
 09/04/2018
 - bypass introspection cache on demand. 
   Sometimes it is necessary to force introspection call for token. If you want to force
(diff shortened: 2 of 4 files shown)


zandbelt:
Actually that use case was supposed to be addressed by setting `opts.introspection_interval`. Doesn't setting that to e.g. `1` or even `0.001` address your needs?

dmitriyblok:
Unfortunately it is not. I'll try to explain my use case. I have two type of calls. Let's call them calls A and B. Majority of the calls are A. They are made by user interacting with Web page. Calls B are made by JS periodically regardless of whether user is doing anything or not. I need calls B to always get introspected regardless of whether token is cached or not. Also, I need them to not cause calls A to always be cached. Because if cache is set to let's say 30s and calls B are made every 15s then calls A will never get introspected. 
Setting opts.introspection_interval to 0.001 will effectively disable cache entirely, which will decrease performance. 

I am using opts.ignore_cache in conjunction with my reverse proxy configuration where I enable/disable cache based on request. Below is a snippet of my nginx.conf.
local opts = {
      .....
        introspection_interval = 30
       ....
      }

if (string.sub(ngx.var.request_uri,-string.len("/sessiontimeout"))=="/sessiontimeout") then
          -- add special parameter to bypass cache
          opts["ignore_cache"] = "yes"
end
 local res, err = require("resty.openidc").introspect(opts)

Hopefully, this explains it.

zandbelt:
I see; if you rename the option to `introspection_cache_ignore` @bodewig and I may consider including it.

dmitriyblok:
Will do

zandbelt:
@dmitriyblok would you also consider avoiding the nested if/then by refactoring the existing code a bit? I believe that makes it an even more trivial addition. @bodewig whaddasay?

dmitriyblok:
Sure. Will do

zandbelt:
Also, feel free to add yourself to the AUTHORS file :-)

dmitriyblok:
Done

bodewig:
Looks fine in general.

I really don't want to play "fetch me a rock", but I'd prefer to make `introspection_cache_ignore` a boolean `true`/`false` rather than a string as we've got several other options that use booleans.

zandbelt:
I felt like I should make the requested changes ;-) completely untested though, so @dmitriyblok please confirm that this works for you and @bodewig if you think it looks good you may merge.

bodewig:
will do once @dmitriyblok confirms it works

dmitriyblok:
Just tested. Works as expected. Thank you!

PR #197, opened 2018-09-14
Avoid key confusion

zandbelt:


bodewig:
you are right, does 2e45631 look better?

PR #208, opened 2018-10-13
url-encode the Client ID for the Windows Integrated Authentication (WIA) case

vishnu2kmohan:
* Conform to https://tools.ietf.org/html/rfc6749#section-2.3.1 for Active Directory Federation Services (AD FS) with WIA
* Minor indentation fix


bodewig:
many thanks!

PR #224, opened 2018-11-05
Added openidc_ensure_discovered_data to openidc_authorization_response

mijohansen:
When running against a Discovery-url I get the following error:

```
2018/11/05 14:30:58 [error] 30#30: *1 lua entry thread aborted: runtime error: /usr/local/openresty/luajit/share/lua/5.1/resty/openidc.lua:1007: attempt to concatenate field 'issuer' (a nil value)
```
When running  `openidc_ensure_discovered_data(opts)` before the check inside the function `openidc_authorization_response` that seem to fix the problem.

### DIFF
--- a/lib/resty/openidc.lua
+++ b/lib/resty/openidc.lua
@@ -995,7 +995,10 @@ local function openidc_authorization_response(opts, session)
     log(ERROR, err)
     return nil, err, session.data.original_url, session
   end
-
+  
+  -- ensure that discovered data is resolved
+  openidc_ensure_discovered_data(opts)
+  
   -- check the iss if returned from the OP
   if args.iss and args.iss ~= opts.discovery.issuer then
     err = "iss from argument: " .. args.iss .. " does not match expected issuer: " .. opts.discovery.issuer


zandbelt:
interesting: I guess you bumped into this because your OP is returning an "iss" parameter in the authorization response, can you confirm that?

Anyhow, a little further down is the `err = ensure_config(opts)` clause that should really be moved up a few lines, up to where you added the `openidc_ensure_discovered_data` call now. (having both is redundant)

mijohansen:
Yes, or OP is returning the "iss" parameter.

zandbelt:
do you want to update the patch or do you want me to do it? (if you you do you get to add yourself to the AUTHORS file... )

zandbelt:
@mijohansen  all set now, thanks; @bodewig as this is an actual bug, do you think we should release 1.7.1? and if so, anything else that you'd want to go in?

bodewig:
We could probably merge #217 right away an I could mix in the `client_secret_jwt` support. But if you want to cut 1.7.1 first, that's also fine with me.

zandbelt:
I'll hold off, the "iss" parameter is a non-core option that not too many OPs use; take your time to prep the additional authentication methods.

PR #229, opened 2018-11-08
Support client_secret_jwt auth method

bodewig:
As a follow-up to #217 this adds support for `client_secret_jwt`.

@pamiel I've unified the two assertions we've built and think I've used all of your logic, it would be good to have a another pair of eyes. Also your code added the `client_id` as an extra parameter to the call, that I didn't have. Reading the spec this seems to be an optional parameter and I added it to both cases as it won't hurt.


pamiel:
Looks good to me; good merge of the 2 inputs !

PR #244, opened 2019-02-05
add Azure AD authorize login_hint parameter

aaalex474:


zandbelt:
this is very OP specific (though optional) so I'm not included to include this in the main branch but more importantly: when and how would this code be triggered (i.e. reauathentciate with existing session)?

aaalex474:
Exactly. The context is reauthentication with existing application session and multiple user accounts.

I found out that it's the only option to make Azure AD to **silently** select the right account if the user is authenticated with multiple accounts. E.g. personal account, work account, particular tenant account and so on. Otherwise the dialog page is displayed or authenticate fails if prompt="none".
Without login_hint it's impossible to refresh silently id_token.

zandbelt:
assuming you use `opts.force_reauthorize` from the calling code you can pass in `opts.authorization_params` with a `login_hint` key from that same location, isn't it?

aaalex474:
Yes, I pass opts.force_reauthorize. But I don't know the right account at this point. It's in the session. And getting/decoding the session myself isn't a good option.

zandbelt:
It is: see samples about obtaining data from the session

aaalex474:
Ok. As you prefer.

PR #254, opened 2019-03-17
Make sure err is a local variable

### LINKED ISSUE #253: Bug with using `revoke_tokens_on_logout = true`
Turning on `revoke_tokens_on_logout` produces the following warning in the logs.

```
_G write guard:12: __newindex(): writing a global lua variable ('err') which may lead to race conditions between concurrent requests, so prefer the use of 'local' variables
stack traceback:
	/usr/local/share/lua/5.1/resty/openidc.lua:1171: in function 'openidc_revoke_token'
	/usr/local/share/lua/5.1/resty/openidc.lua:1213: in function 'openidc_logout'
	/usr/local/share/lua/5.1/resty/openidc.lua:1381: in function 'authenticate'
```

PR to follow

arcivanov:
fixes #253

### DIFF
--- a/lib/resty/openidc.lua
+++ b/lib/resty/openidc.lua
@@ -1159,7 +1159,7 @@ local function openidc_revoke_token(opts, token_type_hint, token)
   end
 
   -- ensure revocation endpoint auth method is properly discovered
-  err = ensure_config(opts)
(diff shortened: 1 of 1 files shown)


bodewig:
Many thanks!

PR #260, opened 2019-04-08
reuse session data during authentication if session data is available

salgadokk:
Hi, I've been using this module in my application and stumbled across a problem.
I thought of using this PR to provide both a description of the problem as well as a possible fix for it.

How to reproduce the problem:
1. In browser window 1 open an application URL which requires authentication.
2. In browser window 2 open an application URL which requires authentication.
3. Return to window 1 and complete the login.
4. "This site can’t be reached" error is displayed on the screen.

The problem is to do with the fact that during authorization, the session data state and nonce always get refreshed.
This causes the following error message in openresty:

`2019/04/08 13:43:15 [error] 7#7: *1 [lua] openidc.lua:1055: authenticate(): state from argument: abcd does not match state restored from session: bcde`

In order to fix this problem, if the session data state and nonce are available (browser window 1 in the example will cause it to be available), then it should not be updated with new values when a new authorization request is performed (browser window 2 from the example).

### DIFF
--- a/lib/resty/openidc.lua
+++ b/lib/resty/openidc.lua
@@ -315,9 +315,20 @@ local function openidc_authorize(opts, session, target_url, prompt)
   local resty_string = require("resty.string")
 
   -- generate state and nonce
-  local state = resty_string.to_hex(resty_random.bytes(16))
-  local nonce = (opts.use_nonce == nil or opts.use_nonce)
-    and resty_string.to_hex(resty_random.bytes(16))
+  local state = nil
+  if (session.data.state == nil) then
+    state = resty_string.to_hex(resty_random.bytes(16))
+  else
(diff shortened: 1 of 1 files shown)


zandbelt:
I'm not so sure about this, it may open a window for attacks i.e. session fixation. In the end parallel requests are not supported, not even after this change.

salgadokk:
Thank you very much for reviewing this PR and for identifying the mistake. You are right, the suggested change is very likely not acceptable.

Do you have a suggestion on how do deal with my problem? The problem with things as they are is that I am getting the "This site can't be reached" error, but the login actually succeeded - if I open again the original url I am logged in. This creates a very bad experience for the user.

What do you mean by "parallel requests are not supported"?

PR #262, opened 2019-04-16
Fix #261: Validate id_token only if we indeed have one

### LINKED ISSUE #261: Pure OAuth2 scenario not supported because of id_token validation
Connecting to a pure OAuth2 server (that is, servers that do not comply with OpenID Connect) is almost working, but lead to a JWT validation error to a missing condition.

###### Expected behaviour

According to file README.md "lua-resty-openidc is a library for NGINX implementing the OpenID Connect Relying Party (RP) and/**or the OAuth 2.0 Resource Server (RS) functionality**." The pure OAuth2 scenario should therefore work without error.

###### Actual behaviour

Execution fails in `openidc_load_jwt_and_verify_crypto` because `id_token` is nil:

    bad argument #1 to 'match' (string expected, got nil)

###### Minimized example

Pure OAuth2 do not provide "discovery" mechanism; JWT based authentication is also generally not supported. It is therefore required to explicitly specify the authorization and token endpoints in `opts.discovery`, and to set option `token_endpoint_auth_method` to either `client_secret_post` or `client_secret_basic`. For example:

            local opts = {
                discovery = {
                    authorization_endpoint = "https://auth.example.net/oauth2/authorize.php",
                    token_endpoint = "https://auth.example.net/oauth2/token.php"
                },

                token_endpoint_auth_method = "client_secret_post",

                -- Might need more options here...

                session_contents = {
                    access_token = true
                }
            }

mjameswh:
Fix #261

### DIFF
--- a/lib/resty/openidc.lua
+++ b/lib/resty/openidc.lua
@@ -1094,9 +1094,12 @@ local function openidc_authorization_response(opts, session)
     return nil, err, session.data.original_url, session
   end
 
-  local id_token, err = openidc_load_and_validate_jwt_id_token(opts, json.id_token, session);
-  if err then
-    return nil, err, session.data.original_url, session
+  local id_token
+  if json.id_token then
+    id_token, err = openidc_load_and_validate_jwt_id_token(opts, json.id_token, session);
+    if err then
+      return nil, err, session.data.original_url, session
+    end
   end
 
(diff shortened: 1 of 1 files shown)


zandbelt:
https://github.com/zmartzone/lua-resty-openidc/issues/261#issuecomment-483841062

PR #285, opened 2019-09-20
Close session

vershnik:
lua-resty-session v2.24 fixed a bug in session locking mechanism, so now storages with locking support can be used in resty

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -32,3 +32,4 @@ reporting bugs, providing fixes, suggesting useful features or other:
 	Oleander Reis <https://github.com/oleeander>
 	Michael Johansen <https://github.com/mijohansen>
 	Joshua Erney <https://github.com/JoshTheGoldfish>
+	Andrei Bastun <https://github.com/vershnik>
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,6 @@
+09/20/2019
+- session started with session.start must be closed in order to release lock
+
 07/15/2019
 - release 1.7.2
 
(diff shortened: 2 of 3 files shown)


bodewig:
The semantics of `close` in https://github.com/bungle/lua-resty-session#boolean-string-sessionclose are a bit vague for my taste.

We return the session to the calling code and it would be perfectly valid for the calling code to use the session object and modify it, would this work for a "closed" session? Or would it work for cookies and not work when using a different backend?

IMHO we have to expect that code exists that calls `authenticate` and then modifies the session and calls `save` once again.

vershnik:
I did more investigations:
calling session:save() without boolean parameter explicitly set to false will close the session.

closed session can be open again (leads to acquiring lock again)
calling save on closed session will save data in a way like it works without using lock (there is parameter to disable locking in each session storage engine)

I agree that caller may want to continue work with session (unless we say that this session is for storing internal data used for openidc and give it unique name using session_opts.name).

Returning closed session may be unexpected in the same way as returning open (and therefore locked) session.

I think there is no other solution than mention chosen behaviour in documentation.
It may look like this (if decided to keep returning open session):


``` 
local res, err, original_url, session = require("resty.openidc").authenticate(opts, nil, "", session_opts)
-- close session to release lock (for cases with session storage with lock support)
session:close();
if err then
    -- handle error
end
```

bodewig:
I agree we've been closing the session under some circumstances in the past via `save` anyway, so I'm fine with your suggested changes and explicitly stating what is going on in the docs. @zandbelt ?

AFAICT we need to update lua-resty-session to at least 2.23 (currently it is >= 2.8) or add guards around the calls to `session:close`.

vershnik:
Another thing to consider - closing session when calling save() happens just before ngx.redirect()
So, calling code does not continue afterwards.

The more I look, the more I think that this pull request can be rejected and we should go with updating documetation with example of closing session after call to authenticate()

bodewig:
OK, agreed. Do you want to replace this PR with a new one updating the documentation?

PR #286, opened 2019-10-01
Refresh failed pass

cretzel:
Problem: When unauth_action is "pass" and refresh of token fails, e.g. because refresh token has expired, then the session attribute "authenticated" is still true. In this case the request will be passed, although the user is not authenticated.

Setting session.data.authenticated to false and returning an error will allow a client to detect, if the refresh was successful.


bodewig:
Thank you @cretzel !

cretzel:
Just my two cents. Thank you for maintaining this project.

Issue #305, opened 2020-01-31
Possible XSS Vulnerability by Returning URI Arguments


barrelmaker97:
I'm using lua-resty-openidc as part of the [kong-oidc](https://github.com/nokia/kong-oidc) plugin and while running ZAP scans against my endpoints behind the plugin, I got results showing that when `<script></script>` tags are included as arguments in the URI, they are sent back to the client, indicating a possible cross-site scripting vulnerability. The results look something like this:
![image](https://user-images.githubusercontent.com/19742024/73579922-84d3dd00-4449-11ea-8161-0fccfc8a00f7.png)After doing some searching, I found that when validating the session state, lua-resty-openidc will return both the provided argument and the session state back to the user if they do not match.
https://github.com/zmartzone/lua-resty-openidc/blob/95b525941cd45be25e81430d01fb8b643a7bbd64/lib/resty/openidc.lua#L1060
Why does this need to be sent back to the user/printed out? Even if the input is properly sanitized, it doesn't seem necessary. If it truly isn't needed, I'm happy to open a PR and change it.

zandbelt:
I believe you're right, there's no need to echo that; a PR would be most welcome, thanks

PR #306, opened 2020-01-31
Do not send back URI arguments to prevent XSS

barrelmaker97:
Fix for #305, no longer sending back arguments provided in the URI or session details.


barrelmaker97:
@zandbelt Fixed, external logs are clean and internal logs are still detailed.

barrelmaker97:
@zandbelt As a side note, is it possible to release a patch version with this change in it? Just wondering if you have a release schedule.

PR #307, opened 2020-02-06
Ability to disable keepalive

3bit-techs:
Hello everyone,
In one of our customers, we use this (amazing) library to behave like a lightweight and fast oidc reverse proxy (deployed as a k8s sidecar along with SPA applications).
To be able to proxy page requests and also ajax requests, we configured the proxy to have two default actions (for non authenticated requests): 

- For ajax requests we deny every request that has failed to **authenticate**
- For page requests we don't specify an action (nil), redirecting to the OP if the requests it's not already logged in

However, after some time we started to notice intermittent failures with the ajax requests. The request failed but the session was valid (so it wasn't because HTTP session expiration, SSO session revoked, etc.).

On the openresty logs we found this error:

`2020/02/06 18:32:57 [error] 8#8: *11162 [lua] openidc.lua:1406: authenticate(): lost access token:accessing token endpoint (https://sso.acme.com/auth/realms/realm/protocol/openid-connect/token) failed: closed, client: x.x.x.x, server: frontend.project.svc, request: "GET / HTTP/1.1", host: "frontend.acme.com", referrer: "https://frontend.acme.com/api"`

After some research we figured out that the lua-resty-openidc uses a library called lua-resty-http to make calls to the OP.

This library (lua-resty-http) by default enables keepalive on the connections so it can create a connection pool to cache and reuse them later. This is necessary for a lot of good reasons (mostly performance related), but for some cases this behaviour can generate this issues, because the connection eventually will be closed and in my case this intermittent errors caused unnecessary api retries and extra traffic.

The idea of this PR is to make it possible for the user to enable or disable keepalive on the lua-resty-http library (just like the ssl_verify option).

The default behaviour would be to keep the keepalive enabled, to avoid causing problems with other users.

Regards,

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -33,3 +33,4 @@ reporting bugs, providing fixes, suggesting useful features or other:
 	Michael Johansen <https://github.com/mijohansen>
 	Joshua Erney <https://github.com/JoshTheGoldfish>
 	Nick Wiedenbrueck <https://github.com/cretzel>
+        Eduardo Gonçalves <https://github.com/Dudssource>
--- a/AUTHORS
+++ b/AUTHORS
@@ -33,4 +33,4 @@ reporting bugs, providing fixes, suggesting useful features or other:
 	Michael Johansen <https://github.com/mijohansen>
 	Joshua Erney <https://github.com/JoshTheGoldfish>
 	Nick Wiedenbrueck <https://github.com/cretzel>
-        Eduardo Gonçalves <https://github.com/Dudssource>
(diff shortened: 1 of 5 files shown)


bodewig:
Looks good to me.

You may want to add something to the changelog and add yourself to the authors file.

3bit-techs:
> Looks good to me.
> 
> You may want to add something to the changelog and add yourself to the authors file.

Done!

Thank you @bodewig .

bodewig:
Thank you!

PR #317, opened 2020-03-26
Fix default expiration-time of jwks and discovery urls caches.

janssen92:
Fix default expiration-time of jwks and discovery urls caches.

The expiration time is measured in 0.001 seconds, according to [https://openresty-reference.readthedocs.io/en/latest/Lua_Nginx_API/#ngxshareddictset](https://openresty-reference.readthedocs.io/en/latest/Lua_Nginx_API/#ngxshareddictset)

Hence, currently jwks and discovery were only cache for 86.4 seconds instead of full day.
=> Replace 24 * 60 * 60 with 24 * 60 * 60 * 1000 in default cache expiration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -380,8 +380,8 @@ lAc5Csj0o5Q+oEhPUAVBIF07m4rd0OvAVPOCQ2NJhQSL1oWASbf+fg==
              -- verified at all.
              --accept_unsupported_alg = true
 
-             -- the expiration time in seconds for jwk cache, default is 1 day.
-             --jwk_expires_in = 24 * 60 * 60
+             -- the expiration time in 0.001 seconds for jwk cache, default is 1 day.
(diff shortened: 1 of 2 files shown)


bodewig:
This is not how I read it. It says

> The optional exptime argument specifies expiration time (in seconds) for the inserted key-value pair. The time resolution is 0.001 seconds

**in seconds** 

My read is you can use a floating point number as expiration time in seconds and you can't expect a higher resolution than millis.

Have you actually verified the cache expires after 86s (I haven't, I'm merely speculating :-) )?

janssen92:
I think you are right. I had caching problems, this seemed to be a reasonable problem, but obviously I didn't read carefully.
My (most likely) problem was the missing lua_shared_dict entries in the nginx server config.

PR #320, opened 2020-03-27
add optional support for PKCE

bodewig:
Initially PKCE has only been used for publci clients but recent security best practices recommend enabling it for confidential clients as well. See also the recent OAuth 2.1 draft https://tools.ietf.org/html/draft-parecki-oauth-v2-1-01


dholth:
I was looking for this today.

Issue #308, opened 2020-02-11
Question: How do I implement front-channel logout?


jnivala:
I am using Lua-resty-openidc as an RP (Open ID Connect Client). My OP is IdentityServer v4.

How do I implement front channel logout? I have multiple applications that authenticate against the OP. When user logs out from any of them, I wish the remaining clients to log out automatically, too.

IdentityServer allows me to specify FrontChannelLogoutUri. Is there a way to configure this in Lua-resty-openid or do I need to implement this by myself?

bodewig:
Right now there is no support for any of the OP initiated logout specs.

Implementing front-channel logout shouldn't be too difficult because all we'd really have to do was to destroy the session if I understand the spec correctly.

Even without lua-resty-openidc supporting it directly you should be able to add a `content_by_lua_block` to a location of your choice which basically does a 

```
    local s, is_existing = require 'resty.session'.open()
    if is_existing then
      s:destroy()
    end
    ngx.header["Cache-Control"] = "no-cache, no-store"
    ngx.header.content_type = "text/html"
    ngx.print('<html></html>')
```

Of course I may be overlooking something. At least we should verify the token we have stored inside of the session has been issued by the `iss` and matches the `sid` given to avoid rogue logouts by third parties.

If we wanted to find the session based on `sid` and not assume we know the current user session via a cookie stored inside of the browser then things may become more difficult.

Also the browser must be willing to send the cookie when loading the iframe which may require proper `SameSite` settings depending on your browser.

jnivala:
Thanks, that worked!

For reference:
Request issuer is escaped. It can be unescaped this way:
`ngx.unescape_uri(ngx.var.arg_iss)`

Identityserver sends lower case url of itself as issuer.

Request sid is here:
`ngx.var.arg_sid`

One should probably compare it against session sid that can be found here:
`s.data.id_token.sid` where s is the session as in bodewig's example.

PR #321, opened 2020-03-27
support OpenID Connect Front Channel Logout

bodewig:
this adds a function that can be used as lua content handler for a frontchannel logout URI.

see also #308

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,9 @@
   endpoint if this has not been done before in order to detect the URI
   of the userinfo endpoint
 
+03/27/2020
+- added a function that implements OpenID Connect Front Channel Logout
+
 02/06/2020
 - ability to disable keepalive from lua-resty-http
   By disabling keepalive we disable the native connection pool,
--- a/README.md
+++ b/README.md
@@ -311,6 +311,43 @@ local res, err, target, session = require("resty.openidc").authenticate(opts)
 session:close()
 ```
 
+## Front-Channel Logout
+
+The `front_channel_logout` function can be used as a content handler
+to implement [OpenID Connect Front-Channel
+Logout](https://openid.net/specs/openid-connect-frontchannel-1_0.html).
+
+```nginx
+  ...
+  location /fc-logout {
+    content_by_lua_block {
+      local opts = {
+        -- session_required = true,
+        -- If this parameter is not set or set to true, requests to
+        -- the front channel logout URI must include the iss and sid
+        -- parameters.
+        -- This also requires id_tokens to be stored inside of the session.
+        -- Default is true
+
+        -- opts.iss = '...',
+        -- if session_required is false the iss parameter can be
+        -- compared to a hard-coded iss value.
+        -- Default is nil
+
+        -- downstream_logout = { 'other-uri', 'yet-another-uri' }
+        -- can contain a table of URIs to be notified via new iframes
+        -- included inside of the response
+        -- Default is no downstream logout URIs at all
+      }
+      require("resty.openidc").front_channel_logout(opts)
+    }
+  }
+```
+
+The `front_channel_logout` accepts as an optional second argument a
+table that is passed to `session.open` in order to read the existing
+session.
+
 ## Sample Configuration for OAuth 2.0 JWT Token Validation
 
 Sample `nginx.conf` configuration for verifying Bearer JWT Access Tokens against a pre-configured secret/key.
--- a/lib/resty/openidc.lua
+++ b/lib/resty/openidc.lua
@@ -1741,4 +1741,62 @@ function openidc.set_logging(new_log, new_levels)
   WARN = new_levels.WARN and new_levels.WARN or ngx.WARN
 end
 
+function openidc.front_channel_logout(opts, session_opts)
+  local session, is_existing = r_session.open(session_opts)
+  local logout_success = true
+  log(DEBUG, 'Front-Channel Logout invoked')
+  if not is_existing then
+    log(WARN, 'no session present, nothing to do')
+    logout_success = false
+  else
+    if not session.started then
+      session:start()
+    end
+    local args = ngx.req.get_uri_args()
+    if not (args.sid and args.iss) then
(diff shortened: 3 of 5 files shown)


eriksunsol:
I'm interested in this as well, so I took some time to investigate...

My understanding is that the whole point of the front channel logout mechanism is that it enables the OP to ask the RP to terminate __all__ active sessions for a user, which is why the `sid` argument can be included in the request. According to the [front channel logout specification section 3](https://openid.net/specs/openid-connect-frontchannel-1_0.html#OPLogout), the OP renders one `<iframe>` tag for each RP which is currently logged in with their respective logout URIs. If I'm right about this, the `sid` argument should not be verified against the id of the current session, but instead the session corresponding to the `sid` argument should be looked up and destroyed.

I'm playing around with AzureAD which supports front channel logout, and in their case (not saying that it is normative) they also support [section 3 of the section management specification](https://openid.net/specs/openid-connect-session-1_0.html#CreatingUpdatingSessions) and provide the `session_state` parameter to the redirect callback. Their interpretation is that the `sid` argument passed to the logout URI matches the `session_state` provided to the redirect callback. The specs are a bit vague so I suppose there may be other interpretations out there.

So to make it work with AzureAD's interpretation of `sid`, you would have to maintain a mapping from `session_state` as received with the redirect callback, to the session, and if `sid` is provided with the logout request it should be used to find the corresponding session to be destroyed. I haven't wrapped my head around the best way to maintain this mapping.

bodewig:
I'm afraid such a mapping depends on how the session is stored. Also I don't think you can look up a session by its id in lua-resty-session, so you cannot destroy any session other than "the current one". Also, there always is only one session attached to a single browser instance (modulo anonymous mode) as at least the session id is stored in a single session cookie.

eriksunsol:
If the session is stored entirely in the cookie, then obviously it is impossible to destroy any other session than the one associated with the browser which is executing the requests, but for the others it should be possible.

I have looked through the source code of [lua-resty-session](https://github.com/bungle/lua-resty-session) and currently exploring the idea to create either a custom session strategy, or a storage adapter which wraps one of the existing ones (except cookie) to maintain the `session_state` mapping and provide a method to wipe all sessions associated with a `session_state` value. Note that especially when using the regenerate session strategy it is likely that there is going to be many sessions in the store associated with a single `session_state` value.

bodewig:
I've thought about a similar approach as well - a wrapper around existing session strategies. This would allow us to support backchannel logout as well.

eriksunsol:
I have a prototype strategy going which which works in my tests with `shm` storage, but should work also with e.g. `redis` or any of the other non-cookie adapters. Actually it's just a small change to make it work with `cookie` too. I decided to re-use the storage adapter already configured for the sessions to store a list of revoked sessions. The strategy has an extended API which allows you to

1. Indicate that a callback (`redirect_uri`) from the OP is being processed and provide the `iss` and `sid` arguments for future revocation. These will be stored in the session.
2. Revoke the session by providing the `iss` and `sid` arguments for sessions to added to the revocation list.

The strategy wraps/inherits another strategy and overrides `save()` to store the identifier for revocation, and `open()` to check the revocation list and set `session.data` to an empty object if the session has been revoked. Ideally only the data owned by __openidc__, or relevant parts of it, should be cleared, so there I have some more work to do.
In order to make it work with `cookie` storage all I need to do is add a configuration option to allow specifying a different storage adapter for the revocation list.

If I can get this to work reliably, extending the concept to support back-channel logout would be easy. There is one aspect of the back-channel logout draft which is problematic though. In a back-channel logout request either a `sub` or a `sid` claim may be included. If the `sid` claim is included the request can be processed similar to a front-channel logout, but if there is just a `sub` claim the revocation list approach is not going to work, and an RP which is also an OP would be unable to reliably propagate the request to downstream RPs. I all other aspects the back-channel draft seems more robust than the front-channel draft and in all honesty I fail to see why they are so different at all. I have posted my suggestions to improve the drafts [here](http://ideas.openid.net/forums/252-ideas/suggestions/40311784-front-channel-and-back-channel-logout-suggestions).

Realizing that I may be hijacking this PR for a broader discussion, please advice if there is a more appropriate place to discuss this.

eriksunsol:
@bodewig I have made a proposal for a revocable session strategy available as PR #330. You are welcome to incorporate it here if you like. You have already put some effort in tests, etc, and I see no reason to duplicate all that work.

zandbelt:
@bodewig do you want me to consider this pull request before releasing, or should I release 1.7.3 as master is right now

bodewig:
Unfortunately my weekends have been devoted to other things lately,

I've been meaning to come back to this to verify @eriksunsol 's comment that I may have misunderstood the spec. In that case merging this PR wouldn't be a good idea, I'm not sure when I will find time to re-read the spec (and my implementation, it's been a while). Let's not hold up the next release and rather be quick with another one once we've sorted things out.

thorstenfleischmann:
I implemented front-channel logout with my IdP using [OIDC Session Management](https://openid.net/specs/openid-connect-session-1_0.html) and the end-session-endpoint. I had to extend "on_authenticated" to save the session_state to the session, so I can use it later (at the front end)

```
    lifecycle = { 
      on_authenticated = function (session)
        -- save session_state to session (for front channel logout)
        session.data.session_state = uri_args.session_state
      end
```

The logout is initiated from the client by calling logout with "Accept: image/png" after receiving a "changed" notification. This may not be suitable for everyone but served my needs.

zandbelt:
FWIW: I think front channel logout is on its way to be deprecated by spec because of recent/upcoming browser changes wrt. 3rd-party cookie handling.

kevinlzw:
Thanks @bodewig for the implementation and everyone who helped work through the logout semantics here. The discussion was valuable, especially around `sid`, current-session logout vs. revoking the OP session, and browser limitations around front-channel logout.

I'm going to close this stale proposal for now. Given the unresolved spec/implementation questions and the age of the PR, a fresh design would be better if logout support is revisited.

PR #330, opened 2020-05-05
Added revocable session strategy to support OP initiated logout scenarios

eriksunsol:
Session strategy to support revocation of sessions for front-channel and back-channel logout scenarios.

This strategy adds the ability to revoke logged in sessions. It will wrap another existing strategy just adding the revocation functionality. To use it, specify `revocable` as the session strategy. It will wrap the `default` strategy by default. To wrap another strategy, set `ngx.var.revocable_session_strategy` to the strategy you want to use, e.g. `regenerate`.

The `revocable` strategy will check sessions against a revocation list. By default the revocation list will be stored using the same storage adapter used by the session. This does not work for the `cookie` storage adapter (for obvious reasons). The storage adapter to be used can be overridden (e.g. if `cookie` is otherwise desired for session storage) by setting `ngx.var.revocable_storage`, e.g. to `redis`. Any additional configuration of the storage adapter can be passed through the `session_opts` argument when calling the **openidc** methods which accept a `session_opts` argument.

To actually revoke a session, call `revocable.revoke()` e.g. from a `content_by_lua_block` directive:

```LUA
content_by_lua_block {
  local session = require 'resty.session'
  local revocable = require 'resty.session.strategies.revocable'

  local openidc_session_opts = {} -- Same as used for openidc calls
  local args = ngx.req.get_uri_args(10)
  local ok, err = revocable.revoke(session.new(openidc_session_opts), args.iss, args.sid)

  if ok then
    ngx.say("Session revoked")
    return ngx.exit(200)
  else
    ngx.say("Revocation failed: " .. err)
    return ngx.exit(500)
  end
}
```

This is just to prove the concept. It would probably be preferred to add the call to the logout processing already called from `openidc.authenticate()` when the `logout_uri` is processed, or if it's more appropriate to add specific **openidc** options to specify front-channel and back-channel logout URIs explicitly.


kevinlzw:
Thanks @eriksunsol for putting this together and for exploring the revocation-list approach. The idea is useful, and the discussion around front-channel/back-channel logout helped clarify what a more complete design would need.

I'm going to close this stale proof-of-concept for now. It was never integrated with lua-resty-openidc logout handling, has no tests or documentation, and targets the old lua-resty-session strategy/storage APIs. If logout support is revisited, it should probably start from a fresh design against the current lua-resty-session API.

PR #350, opened 2020-09-01
Reuse login sessions and ignore following logins

SAGV:
This merge request consists of two small addons intended to improve user experience when having two or more opened login tabs.

---

When `reuse_existing_login_sessions` is set to `true`, the plugin will reuse existing state/nonce which keeps previously opened login tabs valid. 

Currently, all the previously opened login tabs are invalidated as soon as a new tab is opened. This is, for example, an issue when user has multiple tabs with the app opened and is automatically logged out after some inactivity. Then he has a few login tabs but only one has a valid state/nonce. With this fix, all the login tabs are working. 

---

When `ignore_following_logins` is set to `true`, the plugin will ignore successful logins when user is already logged in the app and forward to the app directly, keeping the current login session. 

Currently, if user logins into the app while being logged (for example, by logging in in some old login tab), the plugin returns a error and user is forwarded to a login screen again. With the fix enabled, user is flawlessly redirected back to the application without noticing anything strange. 

---

By default, both parameters are unset so current old behavior stays untouched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,10 +193,18 @@ h2JHukolz9xf6qN61QMLSd83+kwoBr2drp6xg3eGDLIkQCQLrkY=
 
              --use_nonce = false
              -- By default the authorization request includes the
-             -- nonce paramter. You can use this option to disable it
+             -- nonce parameter. You can use this option to disable it
              -- which may be necessary when talking to a broken OpenID
-             -- Connect provider that ignores the paramter as the
+             -- Connect provider that ignores the parameter as the
              -- id_token will be rejected otherwise.
+
+             --reuse_existing_login_sessions = true
+             -- By default a new session and nonce is generated every time authorization is (re)started, invalidating previously open login tabs.
+             -- However, an existing session can be reused so all the previously opened login tabs are valid
+
+             --ignore_following_logins = true
+             -- By default if user logins when already logged, this second login will throw an error because state/nonce we wiped by the first login
+             -- By setting the parameter to true, following logins would be ignored and will redirect user to return url
 
              --revoke_tokens_on_logout = false
              -- When revoke_tokens_on_logout is set to true a logout notifies the authorization server that previously obtained refresh and access tokens are no longer needed. This requires that revocation_endpoint is discoverable.
--- a/README.md
+++ b/README.md
@@ -180,7 +180,9 @@ h2JHukolz9xf6qN61QMLSd83+kwoBr2drp6xg3eGDLIkQCQLrkY=
              --  Expiration leeway for access_token renewal. If this is set, renewal will happen access_token_expires_leeway seconds before the token expiration. This avoids errors in case the access_token just expires when arriving to the OAuth Resource Server.
 
              --force_reauthorize = false
-             -- When force_reauthorize is set to true the authorization flow will be executed even if a token has been cached already
+             -- When force_reauthorize is set to true the authorization flow will be executed even if a token has been cached already. 
+             -- If set, will override `reuse_existing_login_sessions` option.
+
              --session_contents = {id_token=true}
              -- Whitelist of session content to enable. This can be used to reduce the session size.
              -- When not set everything will be included in the session.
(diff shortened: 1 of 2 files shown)


bodewig (changes requested):
This looks mostly fine, but I'm afraid it won't work for the case where a token has expired and we don't own a refresh token. In that case we may start a new authorizattion code grant flow but `session.authenticated` is still true.

Also `reuse_existing_login_sessions` and `force_reauthorize` don't work well with each other. Maybe we just need to document it as in "don't use them at the same time" or we could store we are inside of a forced authentication flow inside of the session as well.

Of course I would have loved tests but realize they'd be somewhat tricky. :-)

SAGV:
@bodewig 

> This looks mostly fine, but I'm afraid it won't work for the case where a token has expired and we don't own a refresh token. In that case we may start a new authorizattion code grant flow but session.authenticated is still true.

Alright, I've added a check for token expiration time. If token already expired — it will not ignore following logins.

> Also reuse_existing_login_sessions and force_reauthorize don't work well with each other. Maybe we just need to document it as in "don't use them at the same time" or we could store we are inside of a forced authentication flow inside of the session as well.

For this one, I marked limitations in readme and will ignore `reuse_existing_login_sessions` whenever `force_reauthorize` is set to true. I am not sure how they might interfere but I still agree that we better not to reuse anything whenever reauth is forced

bodewig (approved):
Technically my concerns have been addressed, so I can approve the change.

Honestly I'm not convinced `ignore_following_logins` works the way you intend it to. When setting it to true the response will be to redirect the user to the last `original_url` stored inside of the sessionpresented right now. There is no guarantee this return_url actually is the place your login tab intceded to go to in the first place. You may be sending people somewhere else entirely. The same is true for `reuse_existing_login_sessions` as you now have two authorizaton code grant flows using the same nonce/state parameters but who actually may have wanted to go to different places.

I believe both options only work if all concurrent login session intend to navigate to the same target, which is probably less common than the more general case (for which the new options may even make things worse by looking like successful flows that end up returning different data than intended).

zandbelt:
@bodewig I agree with your concern, @SAGV what do you think?

SAGV:
Well, the main purpose of `ignore_following_logins` is to prevent redirecting back to login screen and showing error (because second session will never succeed - nonce/state were already removed from cookies). Even if it does not return user to the exact needed url, it is still way better alternative than showing error. Same for `reuse_existing_login_sessions` — the intention is to reduce amount of false errors users gets even though it might not always return to the exact right place.

SAGV:
I'd say I'm going to merge it by end of the day unless you want to continue the discussion

bodewig:
I don't believe you'll be able to merge it yourself :-)

Actually I can think of several situations where returning an error is way better than returning a different resource. One case I'm familiar with is silent reauthentication on a page that wants to load several images. I.e. you load a page and your token expires between loading the page and loading the images contained within - or more commonly today, you've already loaded the page long ago and JavaScript now tries to load some  images and the token has gone. If you are using silent refresh via the frontend rather than a refresh token all your images end up being those that was the target of the first request that succeeded. Similar for AJAX requests suddenly seeing JSON responses for different requests and so on.

To me the options only make sense in a very limited context and is dangerous in others.

SAGV:
Haha, indeed. Merge button is not here :-D

Hmm but this is why we disable it by default and ask the developers to enable only if they want this behavior and understand the consequences. 

In our particular case, the security requirements are similar to bank applications and written in such a way that we do not refresh tokens, but rather prefer users to re-login instead, also having automatically expiring session after X minutes of inactivity. Hence they often get login pages and often in a few tabs at once. So for us your concern is not applicable.

If these changes are really that dangerous in your opinion, I can, of course, simply fork and use the fork but I feel like there might be more people needing this exact approach and that we will be more secure by still using the origin repo (and hence getting all the updates automatically), not our fork in the long term.

SAGV:
By the way, we are already using the fork in our dev envs as a tryout and our testers are way happier with these changes in place.

zandbelt:
I'm on the fence on this one: I agree with the concerns and I don't really like cluttering up the config options and give people a gun to shoot themselves in the foot, on the other hand one can argue that the default stays the same and people would deliberately have to turn this on at their own risk; I'll let @bodewig decide... :-)

bodewig:
In order to address my concerns the documentation of both new options should more strongly state what they do.

So for `reuse_existing_login_sessions` there should be something along the lines of "this will also reuse the return_url of the existing login session so when the flow succeeds the client is redirected to the target of the first flow."

And for `ignore_following_logins` instead of "and will redirect user to return url" something like "and will redirect user to return url of the first login that succeeded".

With this said people would better understand what they'd be signing up for.

bodewig:
I just realized I was wrong about `reuse_existing_login_sessions`, it will not use the one of the first session that succeeds but rather the last one that was started. I'm not sure this is a good idea.

mateodelnorte:
We're experiencing similar issues with users running our application in multiple tabs, where subsequent tabs end up hitting the state mismatch error. 
 
 Is there a recommended fix for this, apart from this not-merged PR?

PR #358, opened 2020-09-20
Added prohibit_redirect option to allow XHR to handle redirects

arcivanov:
Cleanup and optimize requires
Optimize duplicate code related to setting access tokens

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,7 @@
 09/20/2020
+- add `prohibit_redirect` option to improve handling of XHR (#357)
+
+09/10/2020
 - release 1.7.3
 
 08/20/2020
--- a/README.md
+++ b/README.md
@@ -249,6 +249,9 @@ h2JHukolz9xf6qN61QMLSd83+kwoBr2drp6xg3eGDLIkQCQLrkY=
              -- defined in RFC 7636 will be used. The code challenge
              -- method will alwas be S256
 
+             -- prohibit_redirect = false,
+             -- this is a special option for XHR where any attempt to redirect
+             -- will result in an error being returned
           }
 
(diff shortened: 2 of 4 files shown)


arcivanov:
@bodewig

arcivanov:
@bodewig bump

arcivanov:
@bodewig bump?

zandbelt:
`unauth_action = deny` was designed to handle the case of not redirecting: this PR would duplicate the functionality for no good reason as far as I can see

the cleanup and un-duplicating of code is appreciated but would go better in separate PRs

arcivanov:
@zandbelt would you be ok with extending the "unauth_action=deny" to the silent reauthentication as well? I don't really care either way what the option is called.

arcivanov:
I'm talking about this condition:
https://github.com/zmartzone/lua-resty-openidc/pull/358/files#diff-1c5839bc51fa1aa064c7206dce60fb3eR1478
being added here: 
https://github.com/zmartzone/lua-resty-openidc/pull/358/files#diff-1c5839bc51fa1aa064c7206dce60fb3eR1504

bodewig:
`prohibit_redirect` strongly overlaps with `unauth_action = 'deny'`. The only real difference is `prohibit_redirect` disables a premature token refresh if `refresh_session_interval ` is set. If you use `unauth_action='deny'` and `refresh_session_interval = nil` you get the exact same result AFAICT. So I believe the new option is redundant.

From an implementation POV you ignore the value of `prohibt_redirect`, it seems that setting it to `true` or `false` has the same effect.

arcivanov:
Yep, managed to get it to work with those too. Withdrawn.

PR #365, opened 2020-10-20
Update test  Dockerfile image

Darguelles:
The `xenial` image is not accessible anymore, returns 404. Instead, I've changed to `bionic` to execute tests inside docker.


bodewig:
Many thanks @Darguelles

PR #384, opened 2021-05-03
Allow lifecycle handlers to fail, altering the course of request processing

arcivanov:


bodewig:
Looks good, thank you. Please update the documentation on life-cycle hooks of the readme as well. Adding tests would be an extra bonus :-)

arcivanov:
Sorry, forgot to add the docs. Thanks for doing it!

PR #402, opened 2021-10-13
Make revoke_token() accessible

thorstenfleischmann:
New function revoke_token() and revoke_tokens()

Enables the usage of revoke_token() for use cases which needs token revocation without logout.


thorstenfleischmann:
Thanks @bodewig for your feedback!

> could you please add documentation as well? Extra bonus for adding tests :-)

I added a new block to the readme and one (very basic) test. It took me quite some time to get into the design of the tests.

> Also please add yourself to the AUTHORS file.

Done

> While looking through your change I realized the log statements will cause a Lua error if `token_type_hint` is `nil` - and it shares this "quality" with the existing `openidc_revoke_token` function, wich I'll fix now.

I took care of that as well

bodewig:
great, thank you.

PR #405, opened 2021-11-04
#404 Possibilty to pass in a existing session

### LINKED ISSUE #404: Possibilty to pass in a existing session
I am missing an option to pass in an existing session. 

This would solve three issues for me:

1) I also use the session to do some stuff (i.E. csrf-checks). Sometimes I'd like to operate on the session before calling `authenticate`. 

2) `authenticate()` always creates a new session if the session-cookie could not be opened (session.start vs session.open). I want to be able to detect if an existing session could not be opened i.E. because of a invalid signature (wrong session secret). 

3) If the access token refresh is executed a new cookie will be generated (session.regenerate). This can cause issues if there are multiple requests from the same user which require a token refresh:

````
Request 1 with existing session 'A': refreshes access token, regenerate creates a new (authenticated) session 'B'
Request 2 with existing session 'A': refresh is already done, session 'A' is unknown. because of session.start() this will also create a new (unauthenticated) session 'C'
Request 3: will use the last cookie which most likely will be session 'C' (unauthenticated).
````

This issue can not be solved within this library because - by design - the library can not know if a new session should be created (session.start) or an existing one should be used (session.open). However in my use case I know exactly which locations should create a new session and which locations should not.

All three issues can be "workarounded" by opening and closing the session twice (in my code before authenticate and in authenticate). But with a session database and locking mechanisms this seems inefficient.

Therefore I think it would be awesome to have an option to pass in an existing session.

thorstenfleischmann:
Fixes #404

### DIFF
--- a/lib/resty/openidc.lua
+++ b/lib/resty/openidc.lua
@@ -1402,8 +1402,12 @@ local function openidc_get_redirect_uri_path(opts)
   return opts.redirect_uri and openidc_get_path(opts.redirect_uri) or opts.redirect_uri_path
 end
 
+local function is_session(o)
+  return o ~= nil and o.start and type(o.start) == "function"
+end
+
 -- main routine for OpenID Connect user authentication
-function openidc.authenticate(opts, target_url, unauth_action, session_opts)
+function openidc.authenticate(opts, target_url, unauth_action, session_or_opts)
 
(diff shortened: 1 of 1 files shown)


bodewig:
thanks @thorstenfleischmann

PR #445, opened 2022-08-10
fix: missing redirect_uri when trying to refresh

closetool:
Signed-off-by: closetool <c299999999@qq.com>

* Add redirect_uri for refreshing


closetool:
@bodewig @zandbelt plz review

PR #447, opened 2022-08-16
Correct passing of 'symmetric_key' and thus fix HS??? verification

liweitianux:
In function `openidc_load_and_validate_jwt_id_token()`, it was passing the
wrong `opts.client_secret` instead of `opts.symmetric_key` as the HMAC
key, and thus caused signature verification failures for HS??? tokens.

Fix the code to pass the correct `opts.symmetric_key`.

Meanwhile, update the REAME a bit to clarify that the `symmetric_key`
option is directly passed to HMAC code and thus cannot be encoded.

### DIFF
--- a/README.md
+++ b/README.md
@@ -378,7 +378,9 @@ http {
 
           local opts = {
 
-            -- 1. example of a shared secret for HS??? signature verification
+            -- 1. example of a shared secret for HS??? signature verification.
+            -- The key is passed as-is to HMAC and thus may contain binary data
+            -- and cannot be base64 or otherwise encoded.
             --symmetric_key = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
             -- in versions up to 1.6.1 this option's key would have been secret
             -- rather than symmetric_key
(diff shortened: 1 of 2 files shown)


bodewig:
According to https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation digit 8 the client_secret is used to sign the ID Token.

liweitianux:
> According to https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation digit 8 the client_secret is used to sign the ID Token.

@bodewig, thank you for the official reference. I must be mistaken to the whole picture.

I was testing RS256 tokens with Keycloak, and it was needing such a change to make it pass the ID token verification.  Now it seems it may be a bug in Keycloak or I was making some mistakes.

I'll read the specification in more detail tomorrow and check Keycloak again.

Thank you.

liweitianux:
@bodewig, I read the specification and various docs, and I was wrong. Thank you for the support and great work.

bodewig:
you're welcome

PR #449, opened 2022-09-12
added support for registrations endpoint

burnes:
Some OAuth providers provide an endpoint for a registration page instead of a login page. By configuring the `registrations_path` and `registrations_endpoint`, the `openidc_authorize` function will redirect to the configured endpoint including the same parameters as the authorization endpoint and prepare the session.

Can be used e.g. for providing direct Links to Keycloak's registration page.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -36,3 +36,4 @@ reporting bugs, providing fixes, suggesting useful features or other:
 	Eduardo Gonçalves <https://github.com/Dudssource>
 	Thorsten Fleischmann <https://github.com/thorstenfleischmann>
 	Tilmann Hars <https://github.com/usysrc>
+	Bernhard Kuzel <https://github.com/burnes>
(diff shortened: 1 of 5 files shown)


burnes:
@zandbelt Could you please take a look at this?

zandbelt:
I don't see why this should be part of lua-resty-openidc. One could just as well provide the registration link of the provider directly to the user instead of redirecting through the path on the RP, isn't it?

burnes:
@zandbelt To have a working link (and backlink), the session must be created and the state/redirect URL etc. must be appended to the registrations URL. At least in my use-case for Keycloak. If I want to do it outside of the library I have to duplicate a lot of code from the library as all the helper methods are private. Thought this could also help others.

burnes:
We don't use it anymore, as it is not OpenID conform

PR #456, opened 2022-10-28
Fix: Conform to Oauth2.0 RFC for encoding client_id and client_secret

msalar-din:
This is in the context of oidc token introspection

Ran into a very interesting case where one of the client credentials in our case had a `@` which was being converted to `%40` by `lua-resty-openidc` because of the uri escaping. It became a cause of failure for a number of token introspections.

Oauth2 RFC defines client_id and client_secret as strings that can have any printable ASCII character - including @ character.

[According to Oauth 2.0 RFC](https://www.ietf.org/rfc/rfc6749.txt)

`client-id  = *VSCHAR` and `VSCHAR  = %x20-7E` which means any printable ASCII character from ` ` to `~`.
stackoverflow explains: https://stackoverflow.com/questions/983291/purpose-of-x20-x7e-in-regular-expressions

So  uri escaping before base64 encoding will most definitely be unwanted in some cases.

### DIFF
--- a/lib/resty/openidc.lua
+++ b/lib/resty/openidc.lua
@@ -447,11 +447,11 @@ function openidc.call_token_endpoint(opts, endpoint, body, auth, endpoint_name, 
   if auth then
     if auth == "client_secret_basic" then
       if opts.client_secret then
-        headers.Authorization = "Basic " .. b64(ngx.escape_uri(opts.client_id) .. ":" .. ngx.escape_uri(opts.client_secret))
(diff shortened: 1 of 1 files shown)


zandbelt:
the encoding is intended and by spec, see: https://github.com/zmartzone/lua-resty-openidc/blob/master/ChangeLog#L159-L161, your Provider needs to fix its implementation (and certify via https://openid.net/certification)

msalar-din:
Awesome. Thanks @zandbelt

PR #457, opened 2022-11-28
added support ES256&ES512 alg

840963657:


840963657:
lua-resty-jwt support ES256&ES512 alg

bodewig:
thank you @840963657 - one needs at least lua-resty-jwt 0.2.1 for this if I understand the changes over there correctly, right?

Don't we also need to change the code that supports JWT based authentication (lines 466 and 476 at least, which would likely also require something like a `client_ec_private_key` opts setting)? Also I believe the JWK parsing code needs to be extended, it currently only extracts RSA keys.

kevinlzw:
Thanks @840963657 for opening this and for pointing out lua-resty-jwt's ES256/ES512 support.

I'm going to close this stale PR for now rather than keep it open indefinitely. The direction makes sense, but this change only marks `ES*` algorithms as supported and treats them as asymmetric. It does not add the rest of the EC support needed for this to work reliably with real OIDC providers, such as selecting `kty: "EC"` keys from JWKS, converting EC JWK material to a usable public key, updating the lua-resty-jwt dependency requirements, or adding regression coverage for ES-signed ID/access tokens.

If ES algorithm support is revisited, I think it would be best to start from a new PR that:

- updates the minimum lua-resty-jwt version if needed,
- adds EC JWK selection and PEM/key handling for `kty: "EC"`,
- covers ES256/ES512 token validation with tests,
- updates README/ChangeLog where user-visible behavior changes.

Thanks again for raising this.

PR #478, opened 2023-04-28
Migrate lua-resty-session to 4.0.3

### LINKED ISSUE #464: Migrate `lua-resty-session` to v4
I've found this nice PR where the author of `lua-resty-session` bumps the usage of the library to v4.

https://github.com/Kong/kong/pull/10199/files

We've had our fair share of problems with cookies in the v3 version (some of them are in open issues), and while it's hard to trust a codebase that is two weeks old, it's scary not to trust it too. 

What do you think about making the migration? Is that something aligned with the project goals? I could potentially also find time in the next 3 months, if the consensus is that it will work better.

balajiv113:
Fixes #464

### DIFF
--- a/lib/resty/openidc.lua
+++ b/lib/resty/openidc.lua
@@ -351,11 +351,13 @@ local function openidc_authorize(opts, session, target_url, prompt)
   end
 
   -- store state in the session
-  session.data.original_url = target_url
-  session.data.state = state
-  session.data.nonce = nonce
-  session.data.code_verifier = code_verifier
-  session.data.last_authenticated = ngx.time()
+  local session_data = {}
+  session:set_data(session_data)
+  session_data.original_url = target_url
+  session_data.state = state
+  session_data.nonce = nonce
+  session_data.code_verifier = code_verifier
+  session_data.last_authenticated = ngx.time()
 
   if opts.lifecycle and opts.lifecycle.on_created then
     err = opts.lifecycle.on_created(session)
@@ -1061,7 +1063,7 @@ local function openidc_load_and_validate_jwt_id_token(opts, jwt_id_token, sessio
   log(DEBUG, "id_token payload: ", cjson.encode(jwt_obj.payload))
 
   -- validate the id_token contents
-  if openidc_validate_id_token(opts, id_token, session.data.nonce) == false then
+  if openidc_validate_id_token(opts, id_token, session:get_data().nonce) == false then
     err = "id_token validation failed"
     log(ERROR, err)
     return nil, err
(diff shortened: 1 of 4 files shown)


darjeeling:
@bodewig IMHO I think this works well. how about it?
@balajiv113 nice job.

darjeeling:
@balajiv113 I think you missed some point of `session:save`

balajiv113:
@darjeeling could you give more context on the issues that you are facing ??

I ran the current changes with the integration tests that was there. All were passing

lordgreg:
Hi there. I have few crazy days behind me figuring out the issue with "request to the redirect_uri path but there's no session state found". Turns out two versions of lua-resty-session are installed (see more info here https://github.com/revomatico/kong-oidc/issues/34). 

Is there anything open here or can the PR be merged?

brafales:
Any updates on this or ETA would be greatly appreciated. As @lordgreg pointed out, this would fix a compatibility issue with Kong version 3 which is currently preventing us from using the https://github.com/revomatico/kong-oidc plugin.

lukeyeager:
I see a warning when I try to use this:
```
2023/07/13 16:25:57 [warn] 7#7: *1 [lua] _G write guard:12: __newindex(): writing a global Lua variable ('session_present') which may lead to race conditions between concurrent requests, so prefer the use of 'local' variables
stack traceback:
  /usr/local/openresty/luajit/share/lua/5.1/resty/openidc.lua:1461: in function 'authenticate'
  /etc/nginx/lua/utils.lua:33: in function 'do_auth'
  access_by_lua(nginx.conf:99):3: in main chunk, client: REDACTED, server: REDACTED, request: "GET / HTTP/1.1", host: "REDACTED", referrer: "REDACTED"
```

nikita-kuznetsov:
I found a bug with function _is_session_. Session v4 cause the type of session.start being nil. To make function woks properly, need to check _state_ property instead of _start_
I used following implementation.
`
local function is_session(o)
    return o ~= nil and o.state and o.state ~= "closed"
end
`

oldium:
> I see a warning when I try to use this:
> 
> ```
> 2023/07/13 16:25:57 [warn] 7#7: *1 [lua] _G write guard:12: __newindex(): writing a global Lua variable ('session_present') which may lead to race conditions between concurrent requests, so prefer the use of 'local' variables
> stack traceback:
>   /usr/local/openresty/luajit/share/lua/5.1/resty/openidc.lua:1461: in function 'authenticate'
>   /etc/nginx/lua/utils.lua:33: in function 'do_auth'
>   access_by_lua(nginx.conf:99):3: in main chunk, client: REDACTED, server: REDACTED, request: "GET / HTTP/1.1", host: "REDACTED", referrer: "REDACTED"
> ```

You need to add `session_present` to local variables:

```lua
local session, session_present
```

in function `openidc.authenticate`.

oldium:
> I found a bug with function _is_session_. Session v4 cause the type of session.start being nil. To make function woks properly, need to check _state_ property instead of _start_ I used following implementation.
> 
> > local function is_session(o)
> > return o ~= nil and o.state and o.state ~= "closed"
> > end

It would be better to check some actually called function, like `save`:

```lua
local function is_session(o)
  return o ~= nil and o.save and type(o.save) == "function"
end
```

oldium:
Access token refresh cannot work, `session:refresh()` is not the same as previous `session:regenerate()`.

oldium:
I reworked the approach (I used `session:get` and `session:set` to work with session data) and tried to address few missing bits (`session` in `session_or_opts`, token refresh, `is_session` check) [here](https://github.com/oldium/lua-resty-openidc/tree/resty-session-4.x) (_untested_ yet, so take it as work-in-progress). I will create PR as soon as I test it.

bodewig:
Unfortunately I myself completely lack the time to understand what has changed between 3 and 4 and how to migrate this library right now. I'm very grateful for your effort here.

A very cursory glance over lua-resty-session 4.x a few weeks (months?) ago convinced me that migrating would mean more than the changes made initially in this PR and then life interfered.

oldium:
All issues mentioned here are fixed in #489. It uses `session:get` and `session:set` API instead of old-API simulation by calling `session:set_data` and manipulating the passed-in table afterwards.

PR #518, opened 2024-06-10
Logout with client id parameter

patryk-sob:
Logout with client_id parameter is allowed due to openid specification. When id_token_hint is not specified, client_id can be used instead. This change is a solution for situation when session_token is expired and in the logout operation we get the error message "Missing Parameters: id_token_hint error on Session timeout.".


bodewig:
could you please add a new configuration param like `redirect_after_logout_with_id_token_hint` for this and document it? Bonus points for extending https://github.com/zmartzone/lua-resty-openidc/blob/master/tests/spec/logout_spec.lua ;-)

PR #523, opened 2024-08-21
Enable caching of negative introspection responses

boscard:
Right now every single introspection response which has `active` field not set to `true` is just rejected. In situation where there are multiple calls using the same expired/revoked/wrong token response from introspection endpoint could be cached for time set in `exp` field.
This change will allow to do so. By simple configuration in introspection service it will be possible to enable or not (by setting proper value to `exp` field) caching of responses for inactive tokens.

### DIFF
--- a/README.md
+++ b/README.md
@@ -384,6 +384,19 @@ from the cache. In order to avoid cache confusion it is recommended to
 set `opts.cache_segment` to unique strings for each set of related
 locations.
 
+## Caching of negative Introspection responses
+
+By default `introspection` cache will not store negative responses.
+This means that bad actor can potentialy try to exhaust introspection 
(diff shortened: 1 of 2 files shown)


zandbelt:
I see the point about a negative cache but it does open up for resource exhaustion attacks: it should at least be configurable, with the default set to "off" and a warning about the former in the docs

boscard:
@zandbelt You are right. I will try to rework this a bit to update both documentation and configuration options :)

boscard:
@zandbelt  I've updated this PR to address your comments. Please let me know what do you think about it.

boscard:
@zandbelt Is there anything else I have to do to be able to merge this change?

PR #561, opened 2026-06-20
Avoid duplicate introspection client credentials

### LINKED ISSUE #556: introspect() sends client credentials in both Authorization header and POST body when using client_secret_basic
### Description

When `introspection_endpoint_auth_method` is set to `client_secret_basic`, the `introspect()` function sends `client_id` and `client_secret` in **both** the Authorization header and the POST body. This violates [RFC 6749 §2.3.1](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1) and causes strict OAuth 2.0 authorization servers (e.g. Authelia) to reject the request with `invalid_client`.

### Root Cause
In [`lib/resty/openidc.lua`](https://github.com/zmartzone/lua-resty-openidc/blob/v1.8.0/lib/resty/openidc.lua#L1789-L1811), the `introspect()` function **unconditionally** adds credentials to the POST body (lines 1793-1798):
```lua
if opts.client_id then
  body.client_id = opts.client_id
end
if opts.client_secret then
  body.client_secret = opts.client_secret
end
```
Then `call_token_endpoint()` adds them **again** to the Authorization header when auth method is `client_secret_basic` (lines 450-458):
```lua
if auth == "client_secret_basic" then
  headers.Authorization = "Basic " .. b64(ngx.escape_uri(opts.client_id) .. ":" .. ngx.escape_uri(opts.client_secret))
end
```
This differs from other code paths like `openidc_authorization_response()`, which correctly delegates all credential handling to `call_token_endpoint()` without pre-populating the body.
### Expected Behavior
When `introspection_endpoint_auth_method` is `client_secret_basic`, credentials should **only** appear in the Authorization header, not in the POST body.

kevinlzw:
## Summary

Fixes token introspection requests configured with `introspection_endpoint_auth_method = "client_secret_basic"` so client credentials are sent only in the Basic Authorization header, not duplicated in the POST body.

## Root Cause

`introspect()` pre-populated `client_id` and `client_secret` in the request body before delegating to `call_token_endpoint()`. For `client_secret_basic`, `call_token_endpoint()` also adds those credentials to the Authorization header, which violates OAuth 2.0 client authentication expectations for strict authorization servers.

## Changes

- Preserve the legacy body credential behavior when `introspection_endpoint_auth_method` is unset.
- Mark the implicit legacy body credential behavior as deprecated and log a warning directing users to configure `introspection_endpoint_auth_method = "client_secret_post"` explicitly when they want credentials in the POST body.
- Let configured introspection auth methods, including `client_secret_basic`, be handled by `call_token_endpoint()`.
- Add regression coverage ensuring `client_secret_basic` introspection sends a Basic auth header without `client_id` or `client_secret` in the request body.
- Add regression coverage for the deprecation warning on the legacy implicit body credential path.

## Validation

```text
docker build -f tests/Dockerfile . -t lua-resty-openidc/test
```

```text
555 successes / 0 failures / 0 errors / 1 pending
```

The one pending test is the existing upstream pending test in `id_token_validation_spec.lua`.


zandbelt:
Two findings, both about the placement of the new deprecation warning (same log(WARN, …) line, distinct mechanisms):

  - lib/resty/openidc.lua:~2240 — Deprecation WARN is emitted on the resource-server hot path, once per introspection cache miss. introspect()
  is called for every Bearer-token request on an RS. The WARN fires whenever introspection_endpoint_auth_method is nil — which is the
  default/most common config. With introspection_cache_ignore = true, or when the introspection response carries no exp claim (so
  set_cached_introspection never caches it, line 2294), the result is never cached and the warning is logged on every single request, flooding
  error.log at WARN level. A deprecation notice should log once (or be rate-limited / downgraded), not per request on a hot path.
  - lib/resty/openidc.lua:~2240 — The WARN fires even when no client credentials are configured, making it misleading. The warning is logged
  unconditionally when introspection_endpoint_auth_method == nil, before the if opts.client_id / if opts.client_secret checks. A deployment
  doing introspection without client_id/client_secret (public client, or auth supplied via introspection_params/a decorator) is told that
  "introspection client credentials [are being sent] in the POST body" and to switch to client_secret_post — neither of which applies, since
  nothing is being put in the body. Gating the warning on if opts.client_id or opts.client_secret then would make it accurate and also narrow
  the hot-path noise from finding 1.

  Both are addressable together by moving the log(WARN, …) inside the credential check and emitting it at most once.

kevinlzw:
@zandbelt thanks for the careful review. I addressed both points in follow-up commit eb2b2c4:

- the deprecation warning is now emitted only when legacy introspection body auth actually sends client_id or client_secret
- the warning is emitted at most once per worker, so repeated introspection cache misses no longer flood the error log

I also added regression coverage for both cases.

PR #568, opened 2026-06-24
Fix concurrent authorization state handling

### LINKED ISSUE #553: flow fails when we open second tab. owerwritting state in session
When I open 2 tabs  this way I opened first tab then opened second tab after returned first tab and tried to sign in flow is failing cos second tab overwrote state of first tab in session I did some debugging and I think we can make state as table 
  store state in the session
```
  session:set("state", state)
  session:set("nonce", nonce)
  session:set("original_url", target_url)
  session:set("code_verifier", code_verifier)
  session:set("last_authenticated", ngx.time())
```
replace with following   
```
local states = session:get("oidc_states") or {}
  states[state] = {
    nonce = nonce,
    code_verifier = code_verifier,
    created_at = ngx.time()
  }
  session:set("oidc_states", states)
  session:set("original_url", target_url)
  session:set("last_authenticated", ngx.time())
```

kevinlzw:
﻿## Summary
- store pending authorization requests by `state` within the session
- restore nonce, PKCE code verifier, and original redirect URI from the matching authorization state
- keep the legacy single-value session keys visible for lifecycle hook and in-flight session compatibility

## Root cause
Concurrent authorization redirects in the same browser session reused fixed session keys (`state`, `nonce`, `code_verifier`, and `original_url`). Starting a second login flow overwrote the first flow's values, causing the first callback to fail state validation or use the wrong transaction data.

## Validation
- `docker build -f tests/Dockerfile . -t lua-resty-openidc/test`
- `docker run -t --rm lua-resty-openidc/test:latest` (`556 successes / 0 failures / 0 errors / 1 pending`)

Fixes #553

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -30,6 +30,9 @@ unreleased
   with client_jwt_assertion_alg.
 - rebuild client assertion JWTs when retrying token endpoint calls after a
   DPoP nonce challenge.
+- keep pending authorization states separate within the same browser
+  session so concurrent tabs do not overwrite state, nonce, PKCE verifier,
+  or original redirect URI values; see #553.
 
 09/13/204
 - cross-tenant requests are fixed with lua-resty session 4.0.x; closes #526
--- a/ChangeLog
+++ b/ChangeLog
@@ -33,6 +33,9 @@ unreleased
 - keep pending authorization states separate within the same browser
   session so concurrent tabs do not overwrite state, nonce, PKCE verifier,
   or original redirect URI values; see #553.
+- prune stale and excess pending authorization states from the session during
+  authorization requests, configurable with authorization_state_expires_in and
+  authorization_state_max_number; see #553.
 
 09/13/204
 - cross-tenant requests are fixed with lua-resty session 4.0.x; closes #526
--- a/README.md
+++ b/README.md
@@ -213,6 +213,16 @@ h2JHukolz9xf6qN61QMLSd83+kwoBr2drp6xg3eGDLIkQCQLrkY=
              -- Connect provider that ignores the paramter as the
              -- id_token will be rejected otherwise.
 
+             --authorization_state_expires_in = 300
+             -- Maximum age in seconds for pending authorization state stored
+             -- in the session. Expired authorization states are removed when
+             -- a new authorization request is created.
+
+             --authorization_state_max_number = 10
+             -- Maximum number of pending authorization states stored in the
+             -- session. When exceeded, the oldest authorization states are
+             -- removed when a new authorization request is created.
+
              --revoke_tokens_on_logout = false
              -- When revoke_tokens_on_logout is set to true a logout notifies the authorization server that previously obtained refresh and access tokens are no longer needed. This requires that revocation_endpoint is discoverable.
              -- If there is no revocation endpoint supplied or if there are errors on revocation the user will not be notified and the logout process continues normally.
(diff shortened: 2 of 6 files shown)


zandbelt:
Thanks for tackling this — the per-`state` transaction store is the right shape and the callback cleanup (removing the completed state, only nulling the legacy keys when they match) looks correct.

One thing I'd want addressed before merge: **the `authorization_states` table can grow without bound.** Entries are only removed on a *successful* callback, so every abandoned flow — user closes the tab, the OP never redirects back, or a bot repeatedly hits a protected URL with the same session cookie — leaves a permanent entry. For cookie-backed sessions that grows the cookie until it exceeds size limits and the session breaks (the user can no longer log in); for shared-dict sessions it inflates stored session size. The `created_at` field you're storing looks intended to drive expiry, but nothing currently reads it.

I'd suggest pruning at the one place growth happens (`openidc_authorize`), bounding both stale entries (TTL) and burst accumulation (a hard cap):

```lua
-- bound the number of pending authorization states stored in the session:
-- drop ones older than a TTL, then enforce a hard cap by evicting the oldest
local function openidc_prune_authorization_states(opts, authorization_states)
  local max_age    = opts.authorization_state_expires_in or 300
  local max_number = opts.authorization_state_max_number or 10
  local now = ngx.time()

  for s, data in pairs(authorization_states) do
    if not data.created_at or now - data.created_at > max_age then
      authorization_states[s] = nil
    end
  end

  local count = 0
  for _ in pairs(authorization_states) do count = count + 1 end
  while count > max_number do
    local oldest_state, oldest_at
    for s, data in pairs(authorization_states) do
      if not oldest_at or (data.created_at or 0) < oldest_at then
        oldest_state, oldest_at = s, (data.created_at or 0)
      end
    end
    authorization_states[oldest_state] = nil
    count = count - 1
  end
end
```

Called right after inserting the new entry, so the just-created state counts toward the cap but, being newest, is never the one evicted:

```lua
  authorization_states[state] = {
    nonce = nonce,
    code_verifier = code_verifier,
    original_url = target_url,
    created_at = now
  }
  openidc_prune_authorization_states(opts, authorization_states)

  session:set("authorization_states", authorization_states)
```

The TTL (default 300s) is the primary mechanism — an authorization round-trip that hasn't completed in a few minutes is dead. The count cap (default 10) is the DoS backstop for a burst that all lands inside the TTL window; the `while` loop only runs when the cap is exceeded, so the common path is a single cheap `pairs` sweep, and `n` is tiny. Naming follows the existing `opts.access_token_expires_in or 3600` convention. If you'd rather not add a new option, the TTL could reuse the session idle lifetime, or you could ship the cap alone.

Pruning only needs to happen here — the callback already removes the completed state, so reads don't need to prune. Worth a short spec asserting stale/excess entries get dropped, plus a note for the two new options.

(Secondary, non-blocking: genuinely *parallel* in-flight authorize requests can still lose state, since `authorization_states` is read-modified-saved without locking — the last save wins. The fix handles sequentially-opened tabs well; might be worth noting that limitation in the issue.)

kevinlzw:
Hi @zandbelt , Thanks for the detailed review. I addressed this in the latest push:

- added pruning for pending authorization_states when a new authorization request is created
- added authorization_state_expires_in with a default of 300 seconds
- added authorization_state_max_number with a default of 10
- made cap pruning keep the just-created state even when multiple states share the same created_at timestamp
- documented the new options and added a ChangeLog entry
- added regression coverage for stale states, excess states, and same-second cap eviction

Validation passes locally with:

\\\sh
docker build -f tests/Dockerfile . -t lua-resty-openidc/test
\\\

Result: 561 successes / 0 failures / 0 errors / 1 pending.

PR #571, opened 2026-09-02
Propagate Lua-Resty-Session Revocation Store Failures

Hockenba:
## Summary

When `lua-resty-session` 4.2 is configured with cookie storage and a revocation
backend, `session:destroy()` can fail if the store is unreachable while
`revocation_fail_mode` is `closed`. Previously, `openidc_logout` ignored that
error and continued the logout redirect, so the user appeared logged out while
the session cookie remained valid after the store recovered.

This change propagates revocation-store failures from logout and session start
when fail mode is `closed`, so callers can handle the error instead of treating
logout as successful.

- add `openidc_destroy_session` and return destroy errors from `openidc_logout`
  and `authenticate`
- add `openidc_start_session` to surface closed-mode revocation errors from
  `session.start` in `authenticate` and `access_token`
- bump `lua-resty-session` to 4.2.0
- add integration tests for `revocation_fail_mode` `open` and `closed` during
  logout

See [Lua-Resty-Session](https://github.com/bungle/lua-resty-session/pull/209) (completed by me) and [APISIX](https://github.com/apache/apisix/pull/13651) (paused until this is merged).

## Checklist

- [X] Tests pass with `docker build -f tests/Dockerfile . -t lua-resty-openidc/test`
- [X] Tests pass with `docker run -t --rm lua-resty-openidc/test:latest`
- [X] `ChangeLog` is updated for behavior changes, bug fixes, features, or user-visible documentation changes


kevinlzw (changes requested):
Thanks for addressing the fail-closed logout behavior. Before merging, please add coverage for the other newly introduced fail-closed path: a closed-mode revocation read failure during `session.start()`.

The current test only makes `revocation.set` fail after session startup, so it does not exercise the new `unable to check session revocation` branch in `openidc_start_session`, nor the `openidc.access_token()` caller. Please add tests that make `revocation.get` fail and verify that both APIs return the error without authenticating the request.

Also, this logic relies on exact `lua-resty-session` error strings while the rockspec accepts any version `>= 4.2.0`. Please either use a stable upstream error signal or constrain and test the supported dependency range, so a later dependency release cannot silently weaken closed-mode behavior.

PR #575, opened 2026-09-11
fix(introspection): coalesce concurrent cache misses

### LINKED ISSUE #573: Concurrent introspection cache misses cause duplicate endpoint requests
###### Environment

- lua-resty-openidc 1.9.1 / current `master`
- OpenResty with multiple worker processes
- OpenID Connect provider: Keycloak (the behavior is provider-independent)
- Downstream integration: Apache APISIX `openid-connect`

###### Expected behaviour

When introspection caching is enabled and concurrent requests present the same uncached token, one request should call the introspection endpoint and the other requests should wait for and reuse that result.

This coordination should apply only to cache misses. Existing cache hits should remain lock-free, and `introspection_cache_ignore = true` should continue to bypass both caching and cache-miss coordination.

Responses that cannot enter the normal cache, such as an endpoint failure or a successful response without the configured expiry claim, should be shared only with requests waiting for the same in-flight operation. A later independent request should retry introspection normally.

###### Actual behaviour

The current flow performs a cache lookup and then immediately calls the introspection endpoint on a miss. There is no atomic coordination between those operations. If 35 requests for the same token arrive while the first introspection is still in progress, all 35 requests call the identity provider.

This creates a thundering herd precisely when the introspection endpoint is slow or overloaded. An endpoint slowdown lengthens the cache-miss window, causing more duplicate introspection calls and potentially worsening the overload.

Apache APISIX is directly affected. Its `openid-connect` plugin delegates bearer-token introspection to this library, and APISIX already configures an NGINX shared dictionary named `introspection`:

- [APISIX introspection call](https://github.com/apache/apisix/blob/efef0e8413aec9e3c3fb0e9dc8d48cf4c3fa4cb3/apisix/plugins/openid-connect.lua#L1078-L1085)
- [APISIX introspection shared dictionary](https://github.com/apache/apisix/blob/efef0e8413aec9e3c3fb0e9dc8d48cf4c3fa4cb3/apisix/cli/ngx_tpl.lua#L488-L490)

Because APISIX commonly runs multiple NGINX workers, a burst of requests for one cold token can create one introspection request per client request rather than one per gateway instance.

###### Minimized example

The following OpenResty configuration uses a deliberately delayed local introspection endpoint and logs every call:

```nginx
worker_processes 4;

events {
    worker_connections 1024;
}

http {
    lua_package_path '/path/to/lua-resty-openidc/lib/?.lua;;';
    lua_shared_dict discovery 1m;
    lua_shared_dict jwks 1m;
    lua_shared_dict introspection 10m;

    server {
        listen 8080;

        location = /introspection {
            content_by_lua_block {
                ngx.log(ngx.ERR, "INTROSPECTION_CALL")
                ngx.sleep(1)
                ngx.header.content_type = "application/json"
                ngx.say('{"active":true,"exp":4102444800}')
            }
        }

        location = /protected {
            access_by_lua_block {
                local opts = {
                    introspection_endpoint = "http://127.0.0.1:8080/introspection",
                    introspection_endpoint_auth_method = "client_secret_post",
                    client_id = "client",
                    client_secret = "secret",
                    introspection_interval = 60,
                }

                local res, err = require("resty.openidc").introspect(opts)
                if err then
                    ngx.status = 401
                    ngx.say(err)
                    return ngx.exit(401)
                end
            }

            content_by_lua_block {
                ngx.say("ok")
            }
        }
    }
}
```

Send 35 requests concurrently with the same token while the cache is empty:

```sh
seq 1 35 | xargs -P35 -I{} \
  curl -sS -o /dev/null \
  -H 'Authorization: Bearer same-token' \
  'http://127.0.0.1:8080/protected?request={}'

rg -c 'INTROSPECTION_CALL' /path/to/nginx/logs/error.log
```

Expected count:

```text
1
```

Observed count with 1.9.1:

```text
35
```

The same behavior can be reproduced in the library's Docker test environment by delaying the introspection response and issuing concurrent requests from the test client.

###### Configuration and NGINX server log files

The minimized configuration and relevant log marker are included above. No access tokens, client secrets, or provider logs are required to reproduce the issue.

shreemaan-abhishek:
## Summary

- coalesce concurrent introspection cache misses for the same token within an NGINX instance
- share cacheable, non-cacheable, and failed introspection results with requests waiting on the same lock generation
- treat the per-token lock as a bounded lease that must outlive the introspection request
- recheck the cache and fall back to direct introspection when coordination times out or fails
- keep cache hits lock-free and preserve explicit `introspection_cache_ignore` behavior
- document the coordination options and cover concurrency, bypass, failure, timeout, and lease behavior

Closes #573.

## Checklist

- [x] Tests pass with `docker build -f tests/Dockerfile . -t lua-resty-openidc/test`
- [x] Tests pass with `docker run -t --rm lua-resty-openidc/test:latest`
- [x] `ChangeLog` is updated for behavior changes, bug fixes, features, or user-visible documentation changes


shreemaan-abhishek:
@kevinlzw could you please review this PR?

kevinlzw (commented):
Sorry for late, please check below comments.

kevinlzw:
The owner token does not make the release generation-safe because dict:get() and dict:delete() are separate shared-dict operations.
A possible interleaving is:
1. Worker A reads the lock and sees its own owner token.
2. A's lease expires and worker B acquires the same lock key for the next generation.
3. Worker A calls delete() and removes B's lock.
A third worker can then acquire the lock and issue another introspection request while B is still in flight. ngx.shared.DICT does not provide an atomic compare-and-delete primitive, so I do not think this implementation can safely support an owner continuing after its lock lease has expired.
Could we instead treat the lock as a lease, require introspection_lock_exptime to exceed the bounded introspection request duration, and avoid the “outlasted generation” contract? That would make the locking model much easier to reason about.

kevinlzw:
I think cache-miss coalescing should remain a transparent optimization. At the moment, if lock acquisition times out, this returns an error without attempting introspection. With the default five-second timeout, a temporarily slow IdP can therefore turn otherwise valid tokens into authentication failures for all waiting requests.
This changes the existing availability behavior: before this PR, those requests would each perform introspection and could still succeed.
Could lock timeout/failure re-check the cache and then fall back to the existing direct introspection path by default? A strict fail-closed mode could be added later as an explicit opt-in, but it should not be the default behavior.

kevinlzw:
The generation-scoped temporary-result relay is doing a lot of work to share endpoint failures and non-cacheable responses. It adds a fairly complex state machine around a cache-miss optimization and also changes failure/retry behavior.
Would it be reasonable to scope the first version to cacheable successful results only? The standard flow would then be: cache lookup, acquire lease, cache lookup again, introspect once, populate the cache, release. For failed or non-cacheable responses, releasing the lease and preserving the existing direct-introspection behavior would be simpler and more backward-compatible.

shreemaan-abhishek:
For the temporary-result relay, the motivating APISIX case is an overloaded introspection endpoint returning 503. Limiting coordination to cacheable successes means every waiter eventually retries introspection, so the overload scenario remains unresolved. Would you be open to retaining in-flight sharing for failed and non-cacheable results after the lease and fallback changes, or would you prefer this PR to address only cacheable successes and defer that behavior?

kevinlzw:
Thanks for the update. The fallback change addresses my main compatibility concern. Your explanation makes sense: limiting coordination to cacheable successes would leave the original 503 overload scenario unresolved, so I’m happy to retain in-flight sharing for failed and non-cacheable results.

A few focused changes would still help:

- Let waiters consume the result of the generation they originally observed without first acquiring the lock or switching to a newer generation.
- Restore the owner check before releasing the lock as a defensive measure, while documenting that it is not atomic and does not fully eliminate the race.
- Add multi-worker tests covering lease expiry and coordination-state eviction. Duplicate upstream calls may occur when coordination breaks down, but coordination failure alone should not cause authentication failures.

I’m comfortable treating coalescing as a best-effort optimization with these limitations documented. Happy to discuss the trade-offs further.

